PR #3, opened 2024-10-04
docs(README): Add a utility script to extract EPFL-recognized association data

oskar-codes:
The script should be executed in the console of the developer tools after navigating to https://www.epfl.ch/campus/associations/list/. The extracted data is available in associations.json.


Zafouche (approved):
LGTM.

Zafouche:
LGTM as well.

PR #10, opened 2024-10-04
Chore/association model

oskar-codes:
This pull request aims to integrate model classes for Associations, namely:
- Data class for Associations
- Data class for Users
- AssociationRepository interface
- AssociationRepositoryFirestore implementing the aforementionned interface using Firestore

Two tests, respectively for the AssociationRepository and the User, are included.


Aurelien9Code (approved):
This look very good to me for now. We might just add in a future a 'parent' association and a 'child' association, both set to Null at first. So that small commissions could join the project before their parent association, and be linked to it after. And so that big associations could register at once all their commissions (if they are not already in the project).

Aurelien9Code (approved):

AlouchLaBouche (approved):
Change the Association data class to better define `val members : List<String>`.

oskar-codes:
@AlouchLaBouche Indeed, in a later pull request I will add a computed property that is automatically populated with the complete member data.

leagrieder (commented):
### PR Review

This is an excellent PR! Here are some highlights:

- **Description:** Clear and to the point, making it easy to understand the changes and their context.
- **Tests:** Your tests are thorough and well-structured, covering all relevant scenarios.
- **Commit Messages:** Consistently great, providing clarity and context for each change.

Overall, a well-executed PR. Great work!

PR #11, opened 2024-10-05
chore(app): change folder structure for better organization

Zafouche:
Just moving files around and keeping an organized structure such that we can keep a logical folder structure moving forward.


Zafouche:
Will do this for the unit tests folder as well.

oskar-codes (approved):
LGTM.

PR #12, opened 2024-10-05
chore(model/association): Create view model for the association database.

Zafouche:
The explore view model will enable us to get the list of associations from the firebase database. This will be useful for screens such as the explore screen, which will display all associations, sorted vertically by category, and horizontally by some other logic (or none).

### DIFF
--- a/app/src/main/java/com/android/unio/model/association/AssociationRepository.kt
+++ b/app/src/main/java/com/android/unio/model/association/AssociationRepository.kt
@@ -1,5 +1,7 @@
 package com.android.unio.model.association
 
 interface AssociationRepository {
+  fun init(onSuccess: () -> Unit)
+
   fun getAssociations(onSuccess: (List<Association>) -> Unit, onFailure: (Exception) -> Unit)
 }
--- a/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
@@ -1,9 +1,19 @@
 package com.android.unio.model.association
 
+import com.google.firebase.Firebase
+import com.google.firebase.auth.auth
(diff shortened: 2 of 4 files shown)


oskar-codes (commented):
Is looking overall good to me. Might need additional tests. Also might need a caching system in the future to prevent excessive requests to Firestore and enable offline features.

oskar-codes (approved):

Zafouche:
@oskar-codes What additional tests would you have in mind?

> Also might need a caching system in the future to prevent excessive requests to Firestore and enable offline features.

Agreed. Will be added when the time comes. Still need to decide what and how we store information about associations (importantly, how we determine an association's category: at sign-up or as a maintenance task that we would have to approve for every new association that signs up.)

oskar-codes:
@Zafouche Maybe test the `init` method of `AssociationRepositoryFirestore`, and the factory of the `ExploreViewModel` just to increase new code test coverage ? I can work on that if needed.

About the assignment of associations category, I believe having the choice at association sign-up is the better option.

Zafouche:
@oskar-codes Have a look at the last test `testInitFetchesAssociations `. I think it tests it well or do you think it is not appropriately done?

leagrieder (commented):
### PR Review

- **Null Handling in `AssociationRepositoryFirestore.init()`:** Currently, the `init()` function assumes that there will always be a valid user after authentication. However, there could be cases where `currentUser` is `null`, and it would be worth handling this case explicitly to prevent potential `NullPointerException` issues. Adding a check would make the code more robust.
- **Description Improvement:** Your description could benefit from being more detailed. You can refer to this [PR template](https://github.com/thephpleague/skeleton/blob/master/PULL_REQUEST_TEMPLATE.md) for guidance on structuring it effectively.
- **Testing:** Great job on thoroughly testing your code! Keep up the good work. 

Well done overall!

PR #14, opened 2024-10-05
feat(explore): implement base of the explore screen

Zafouche:
Things that still need to be implemented:

- [ ] Interactions
        - [ ]  Clicking on an association brings us to their association profile (with a back button available).
        - [ ]  Typing in the search bar should display all accordingly filtered associations, in a simple column lazy list, where the items will show up in a small height but take the screen's width (just like common search suggestions in most applications that we use).
        - [ ] Bottom navigation bar needs to be added.

- [ ] Data queries (can be done later in a different PR)
        - [ ] The explore view model should be used, and use its list that gets the data directly form the firebase database.
        - [ ] A way to fetch their logo and display them instead of the gray box.

- [ ] Design (can be done later in a different PR)
        - [ ] Actual design that's more accurate to the figma when it is finalized. This includes positioning, fonts and colors.

### DIFF
--- a/app/src/main/java/com/android/unio/ui/explore/Explore.kt
+++ b/app/src/main/java/com/android/unio/ui/explore/Explore.kt
@@ -1,0 +1,203 @@
+package com.android.unio.ui.explore
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyRow
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Icon
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Text
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Search
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import com.android.unio.model.association.Association
+
+@Composable
+fun ExploreScreen() {
+    Scaffold(
+        modifier = Modifier.testTag("exploreScreen"),
+        /**
+         * Here, we should have the bottom navigation bar.
+         */
+        content = { padding ->
+            Column(
+                modifier = Modifier
+                    .fillMaxSize()
+                    .padding(padding)
+            ) {
+                OutlinedTextField(
+                    value = "",
+                    onValueChange = {},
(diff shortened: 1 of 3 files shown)


Zafouche:
This is what it looks like for now, none of the actual design is here since the figma needs to be further developed, only the logic is implemented.

Screenshot 2024-10-05 at 19 25 50

**Logic:**
Associations are sorted as such: vertically, they are separated by category, and then in each category, they are displayed and sorted alphabetically (regarding their acronym's first letter) horizontally

Zafouche:
The duplicate code analysis is the mock list of associations, would probably be good to move it on the firebase database when we decide how and when a category of an association is chosen.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
```diff
@@ -0,0 +1,178 @@
...
+            modifier = Modifier.fillMaxWidth(),
+            maxLines = 1)
+      }
+}
+
+/**
+ * Everything below is mock data. This file (and other classes) should later be adapted to get data
+ * from the association repository.
+ */
+enum class AssociationType {
```
Romainhir, line 106: Should probably be in another file even if it is only mock data

Romainhir (approved):

app/src/main/java/com/android/unio/ui/explore/Explore.kt
```diff
@@ -0,0 +1,178 @@
...
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.dp
+import com.android.unio.model.association.Association
+
+@Composable
+fun ExploreScreen() {
+  Scaffold(
+      modifier = Modifier.testTag("exploreScreen"),
+      /** Here, we should have the bottom navigation bar. */
```
oskar-codes, line 35: We should maybe have the `ExploreScreen` composable represent only the Explore content (without the bottom bar), and the actual BottomBar is at a higher abstraction level to prevent code duplication.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
```diff
@@ -0,0 +1,178 @@
...
+                AssociationType.entries.forEach { category ->
+                  val filteredAssociations =
+                      mockAssociations
+                          .filter { it.type == category }
+                          .sortedBy { it.association.acronym }
+
+                  if (filteredAssociations.isNotEmpty()) {
+                    item {
+                      Text(
+                          text = category.name.lowercase().replaceFirstChar { it.uppercase() },
```
oskar-codes, line 57: For code clarity, this could be extracted in its own method. This is of course not an urgent fix, but it is best to keep the logic in a separate view model, or simply at the start of the composable function.

Zafouche: Will do.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
Zafouche: Sure, also would make it more readable. Will be done when the navigation system is available. Or we can merge this and refactor later today.

Zafouche: Done.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
Zafouche: Done.

Zafouche:
Will have to remove the MockAssociations class later in a different PR to remove code duplicates.

leagrieder (commented):
### PR Review

- **Description Improvement:** The description is better than in your previous commits, but it’s important to explain what you’ve done, how you’ve done it, and why, before mentioning what still needs to be done. This helps provide more context to the changes.
- **UI Implementation:** Good job on implementing this UI! It looks great, and the code coverage is excellent as well—well done!

PR #26, opened 2024-10-08
chore(Associations): Adapt association data structure

oskar-codes:
The updated data structure includes a `url` field and default values. The scraping script has also been updated. This is a minor change.


Zafouche (approved):
LGTM. We should all a category field later on.

PR #28, opened 2024-10-08
chore(Association): Add URL field to association hydration

oskar-codes:
Minor fix to the association hydration process, including the association url.


Zafouche (approved):
LGTM. (small issue)

PR #30, opened 2024-10-08
Fix/navigation

Romainhir:
Fix bug with navigation and the main activity

### DIFF
--- a/app/src/main/java/com/android/unio/MainActivity.kt
+++ b/app/src/main/java/com/android/unio/MainActivity.kt
@@ -16,6 +16,8 @@ import com.android.unio.ui.home.HomeScreen
 import com.android.unio.ui.navigation.NavigationAction
 import com.android.unio.ui.navigation.Route
 import com.android.unio.ui.navigation.Screen
+import com.android.unio.ui.saved.SavedScreen
+import com.android.unio.ui.user.UserProfileScreen
 
 class MainActivity : ComponentActivity() {
   override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -33,11 +32,18 @@ fun UnioApp() {
...
+    //      navigation(startDestination = Screen.HOME, route = Route.HOME) {
+    //        composable(Screen.HOME) { HomeScreen(navigationActions) }
+    //      }
+    //      navigation(startDestination = Screen.EXPLORE, route = Route.EXPLORE) {
+    //        composable(Screen.EXPLORE) { ExploreScreen(navigationActions) }
+    //      }
+    //  }
+    navigation(startDestination = Screen.HOME, route = Route.HOME) {
+      composable(Screen.HOME) { HomeScreen(navigationActions) }
+
```
AlouchLaBouche, line 46: Implement here the different screens `Account` and `Saved`

AlouchLaBouche (changes requested):
The app crashes when we try to navigate to the `Saved` and `Account` screens. This is because they are not implemented in the `NavHost`.
Futhermore, the `Map` button should not be in the `BottomNavigationMenu` But should be a `Floating Action Button`. (refer to the [Figma](https://www.figma.com/design/rlvqRyIsbdSyEbXu8Xp7yG/Unio?node-id=0-1&node-type=canvas&t=LvJdEkDqP8Gs3sbG-0))

app/src/main/java/com/android/unio/MainActivity.kt
oskar-codes: I just worked on this and it should be fixed by now.

oskar-codes (approved):

oskar-codes (approved):
The fix is working properly, the unit/connected tests pass locally and on the CI, and the application runs correctly on its own.

AlouchLaBouche (approved):

leagrieder (commented):
### PR Review

Good fix to improve the navigation system, well done, this is a good PR

PR #39, opened 2024-10-09
feat(explore): add association interaction to go to the association profile screen

Zafouche:
Ultimately, some things had to be changed for this to be possible:

1. Modified the navigation system, such that it correctly handles and recognizes screen hierarchy (reminder: only the parent screen should have a route)
2. Added the interaction to the association items to bring us to the association profile screen. Ultimately, in the future, we will have to keep a state as to which association was selected such that we know which association's profile we display.
3. Added a "go back" button to the association profile screen as it is a child screen of the explore screen.

### DIFF
--- a/app/src/androidTest/java/com/android/unio/ui/association/AssociationProfileTest.kt
+++ b/app/src/androidTest/java/com/android/unio/ui/association/AssociationProfileTest.kt
@@ -14,38 +14,36 @@ import org.mockito.kotlin.mock
 import org.mockito.kotlin.verify
 
 class AssociationProfileTest {
-    private lateinit var navHostController: NavHostController
-    private lateinit var navigationAction: NavigationAction
+  private lateinit var navHostController: NavHostController
+  private lateinit var navigationAction: NavigationAction
 
-    @get:Rule
-    val composeTestRule = createComposeRule()
+  @get:Rule val composeTestRule = createComposeRule()
 
-    @Before
-    fun setUp() {
-        navHostController = mock { NavHostController::class.java }
-        navigationAction = NavigationAction(navHostController)
-    }
+  @Before
+  fun setUp() {
+    navHostController = mock { NavHostController::class.java }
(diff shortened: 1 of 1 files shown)


Zafouche:
Will add tests tomorrow morning.

Zafouche:
Will add tests for the explore screen and the navigation system in different PRs. However I don't see how to improve the code coverage on AssociationProfile.kt.

oskar-codes (approved):
Overall, this looks good to me. Can be merged.

Zafouche:
We will do this in sprint 2 in a different PR.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -0,0 +1,40 @@
...
+                  onClick = { navigationAction.goBack() },
+                  modifier = Modifier.testTag("goBackButton")) {
+                    Icon(
+                        imageVector = Icons.AutoMirrored.Outlined.ArrowBack,
+                        contentDescription = "Go back")
+                  }
+            })
+      },
+      modifier = Modifier.testTag("AssociationScreen")) {
+        Text("Association screen")
```
leagrieder, line 38: It's a good habit to develop to avoid writing hard coded string in the code, and favor the use of String.xml resources. This will makes it easier to later [localize](https://developer.android.com/guide/topics/resources/localization) the application. It also makes the app more maintainable as it is easy to modify one text, without having to remember to modify every places that are displaying the hard coded text.

leagrieder (commented):
### PR Review

- **Hardcoded Strings:** There are hardcoded strings in the file (e.g., "Association Profile"). These should be moved to string resources to improve maintainability and localization.
- **Feature Definition:** While it’s good to mention what needs to be changed to implement your feature, it's important to first clearly define what you are adding to the app and how you plan to achieve it.

Good start overall, just focus more on clarity and best practices!

PR #43, opened 2024-10-10
chore(explore): change height of association item for demo

Zafouche:
This is simply to show, during the S1 demo, that we can scroll vertically through the association categories.
The height increase will allow the lazy list to have enough height such that we have to scroll so see the last category.


oskar-codes (approved):
LGTM.

Zafouche:
> LGTM.

Thanks 👍

PR #44, opened 2024-10-10
feature/firestore-images: Implement image repository

armouldr:
Implemented the image repository, i.e a way to upload an image to our Firebase storage and then to get the stored image's download url.


oskar-codes (approved):
Great work Arnaud! This looks all good to me.

armouldr:
Thank you for the feedback ! I will write more explicit PR descriptions from now on. Indeed, a deleteImage feature will certainly be useful in the future, I will add it as a small task.

PR #46, opened 2024-10-11
fix(model/image): upload image

armouldr:
Fix uploadImage so it outputs the image's download link automatically.
Set the getDownloadUrl as private because it shouldn't be used elsewhere in the project.
Also added some comments to the methods.

### DIFF
--- a/app/src/main/java/com/android/unio/model/image/ImageRepositoryFirebaseStorage.kt
+++ b/app/src/main/java/com/android/unio/model/image/ImageRepositoryFirebaseStorage.kt
@@ -1,7 +1,6 @@
 package com.android.unio.model.image
 
 import com.google.firebase.storage.FirebaseStorage
-import com.google.firebase.storage.UploadTask.TaskSnapshot
 import java.io.InputStream
 
 class ImageRepositoryFirebaseStorage(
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/unio/model/image/ImageRepositoryFirebaseStorage.kt
Aurelien9Code, line 1: Very useful changes ! Bravo !

Maybe you could just create a more precise error handling, for example in the getImageUrl method show which issue made the download URL fail, and the same when uploading image 
Example : `onFailure(Exception("Image upload failed because of : ${exception.message}"))`

This might seem unuseful now, but it might be important when implementing strong log system or launching the app : )

Aurelien9Code (approved):
Overall that's really good and the upload process is more user-friendly : ) (so good for us haha) 

As I said, you might change your event handling now, or we will do it all at once on another task
Thanks for your work !

app/src/main/java/com/android/unio/model/image/ImageRepositoryFirebaseStorage.kt
armouldr: Correct me if I'm wrong but I think that the error handling is supposed to be done inside the onFailure method, received as argument by the 2 functions.

Aurelien9Code: I can surely be wrong, but wouldn't it be in app/src/main/java/com/android/unio/model/image/ImageRepositoryFirebaseStorage.kt, directly on the listener ? 

update : sure, caller can handle errors
When i saw it on your code this afternoon, I was thinking that a separation of the different errors could be great directly from the repository
But i think you are right now, if the caller want to have different error handling management

PR #50, opened 2024-10-11
chore(association): refactor association repository with universal function

Zafouche:
This update enhances the quality of our code:
- It can still handles the simple functions like `addAssociation`, `updateAssociation` and `deleteAssociation`, that use a simple `() -> Unit` typed `onSuccess` function.
- It now supports functions where their `onSuccess` function might require a different input type. This includes the `getAssociations` and `getAssociationWithId` functions.

### DIFF
--- a/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
@@ -26,18 +26,19 @@ class AssociationRepositoryFirestore(private val db: FirebaseFirestore) : Associ
       onSuccess: (List<Association>) -> Unit,
       onFailure: (Exception) -> Unit
   ) {
-    db.collection(ASSOCIATION_PATH)
-        .get()
-        .addOnSuccessListener { result ->
-          val associations = mutableListOf<Association>()
-          for (document in result) {
-            val association = hydrate(document)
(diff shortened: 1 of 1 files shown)


Zafouche:
I am curious as to why SonarCloud's quality gate passed. Doesn't it require 80% coverage on new code lines? I assume it's simply because the new lines are code are just refactoring of the old code, which was already well enough tested.

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
Aurelien9Code, line 1: Looks perfect to me ! Well done !

In the end, we will need to find a way to notify the user when an error occured

Aurelien9Code:
> I am curious as to why SonarCloud's quality gate passed. Doesn't it require 80% coverage on new code lines? I assume it's simply because the addition of new lines of code is just due to the refactoring of the old code, which was already well enough tested.

Yes, according to https://docs.sonarsource.com/sonarcloud/improving/quality-gates/, if the majority of changes involved refactoring existing lines (as was the case here), then the newly added lines may not be subject to the same coverage requirement ! : )

Aurelien9Code (approved):
As said, looks perfect !

oskar-codes (approved):
This is a much needed and appreciated change. Way to go 💪

PR #61, opened 2024-10-13
chore(association/profile): implement passage of selected association's uid to the association profile screen  

Zafouche:
## Description

- Implemented the receiving side of the argument passed through the Jetpack Navigation in `MainActivity.kt`.
- Implemented the sending side of the passage of argument in the explore screen `Explore.kt`.
- Added a helper function named `findAssociationById` in `AssociationViewModel.kt`, which fetches the Association object if it is present in the list of associations.
- Added the `AssociationProfileScaffold` composable in `AssociationProfile.kt`, which allows us to either display and error `Text` if the association is not found, either the data (for now just the acronym) of the association in the same scaffold.

## Motivation

This PR implements the method of passing the selected association's uid to the association profile screen using Jetpack Navigation arguments. This approach follows what's explained in this [article](https://sagar0-0.medium.com/pass-arguments-to-destinations-jetpack-compose-navigation-7c29d4c73ce5).
This allows us to get the data of the association profile that has to be displayed on the association profile screen (`AssociationProfile.kt`)

### Considerations

- Helper function in `AssociationViewModel`: I wrote a helper function within the view model to fetch the data with the association's uid. However, I am not sure about this approach. Say the user refreshed the page, we would need to re-call the function and convert the `val association` to a `var`. This feels unnecessary as it would change automatically if we used a state flow in the view model.
- Bootcamp approach: I'm considering the usage of the method that we implemented in the bootcamp, e.g. we have a function that selects a ToDo from the `MutableStateFlow` list present in its view model (which we have in our association view model), and the data can be fetched through another `MutableStateFlow`, where `selectToDo` stores the item from the `MutableStateFlow` list. Now I am not sure about it as I need to do research on it, but would the `selectedToDo` `MutableStateFlow` update itself if the `MutableStateFlow` list is updated? This could happen in the scenario of when the user was offline and gains back connectivity. This can seem far fetched, but the automatic update could be useful.
- Advantage of the usage of argument: If there is an event displayed on the profile, an event that is co-organized by multiple associations, and if we decide to make the small association logos clickable, meaning it would send us to their profile, passing the uid by argument might be best. This is because it would only get the data when the navigation system initiates the navigation to the new screen, however if we have the selection of the association like the selection of a ToDo in the bootcamp, first you would select the ToDo, then navigate. The selection might force the current composable to be redrawn (because it is a state flow), and then navigate to the new one, where it would be yet again redrawn. I am not sure about this, I need to do some research on it, but if this is the case, then it is not something we would want.
Edit: From what I understand in the documentation, this shouldn't happen so this isn't really an argument.

### Note
You cannot yet play around with it, as the view model is not yet used in the explore screen. I will work on using it on a different PR, and come back to this one, pull the changes on this branch so we can test this by hand as well. Tests are also not yet written as I am not sure this method is better than the one used in the bootcamp.


Zafouche:
Other tests have not been adapted as I need opinions on this.

app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -46,7 +51,14 @@ fun UnioApp() {
     }
     navigation(startDestination = Screen.EXPLORE, route = Route.EXPLORE) {
       composable(Screen.EXPLORE) { ExploreScreen(navigationActions) }
-      composable(Screen.ASSOCIATION) { AssociationProfile(navigationActions) }
+      composable(Screen.ASSOCIATION_PROFILE + "/{uid}") { navBackStackEntry ->
+        val uid = navBackStackEntry.arguments?.getString("uid")
+        uid?.let { AssociationProfile(navigationAction = navigationActions, associationId = it) }
+            ?: run {
+              Log.e("AssociationProfile", "Association id is null")
+              Toast.makeText(context, "Association id is null", Toast.LENGTH_SHORT).show()
```
oskar-codes, line 59: This is very minor, but consider using UID instead of ID for consistence, and also written capitalized in the `Toast`.

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
oskar-codes, line 1: Good call on renaming the `ExploreViewModel` to `AssociationViewModel`, it is much more representative.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -11,13 +11,28 @@ import androidx.compose.material3.Text
...
 @Composable
-fun AssociationProfile(navigationAction: NavigationAction) {
+fun AssociationProfile(
+    navigationAction: NavigationAction,
+    associationId: String,
+    associationViewModel: AssociationViewModel = viewModel(factory = AssociationViewModel.Factory)
+) {
+  val association =
+      associationViewModel.findAssociationById(associationId)
+          ?: return Text(
```
oskar-codes, line 31: Good error handling, maybe this could even be split in an `if` statement for code clarity and an additional `Log.e` call.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
```diff
@@ -120,23 +120,29 @@ fun ExploreScreenContent(padding: PaddingValues, navigationAction: NavigationAct
...
-    Image(
-        painter = painterResource(id = R.drawable.adec),
-        contentDescription = "image description",
-        modifier = Modifier.size(124.dp))
+  Column(
+      modifier =
+          Modifier.clickable {
+            navigationAction.navigateTo(
+                Screen.ASSOCIATION_PROFILE +
+                    "/{uid}".replace(oldValue = "{uid}", newValue = association.uid))
```
oskar-codes, line 128: I'd consider extracting the string replacement logic in a separate function to reduce cognitive complexity and enable reusability, as navigation route parameters will likely be used in other parts of the project.

oskar-codes (approved):
Overall very strong code. About the approach used, in my opinion passing a UID as a route parameter is the more scalable solution, but we can discuss this with the team next meeting.

app/src/main/java/com/android/unio/ui/explore/Explore.kt
oskar-codes: Also, I'm noticing this could be simplified to just `"${Screen.ASSOCIATION_PROFILE}/${association.uid}"`

app/src/main/java/com/android/unio/MainActivity.kt
Zafouche: > This is very minor, but consider using UID instead of ID for consistence, and also written capitalized in the `Toast`.

Good catch I will change it.

> Also, this might be debatable but you could integrate the `+ "/{uid}"` directly in the `Screen.ASSOCIATION_PROFILE` constant.

Yes but I would argue that it doesn't fall into the category of a screen name, hence I wouldn't include it in the `Screen` enum. However, I think having an additional enum type for the screen arguments would be useful, don't you think?

app/src/main/java/com/android/unio/ui/explore/Explore.kt
Zafouche: > I'd consider extracting the string replacement logic in a separate function to reduce cognitive complexity and enable reusability, as navigation route parameters will likely be used in other parts of the project.

I would argue that having a separate function is really useful, as I don't see how to both generalize it and make it easy to the eye in terms of what arguments would be needed for the function, since it is possible that other parts of the project would use multiple arguments. 
Maybe I don't see it how you do, how would you define a helper function for this replacement?

> Also, I'm noticing this could be simplified to just `"${Screen.ASSOCIATION_PROFILE}/${association.uid}"`

That's true!

app/src/main/java/com/android/unio/MainActivity.kt
oskar-codes: I see. Maybe we could have something like this:
```kotlin
enum class Screen(private val route: String) {
  HOME("home")
  ...
  ASSOCIATION("association/{uid}");

  fun withParams(vararg params: String): String {
    return params.fold(route) { acc, param -> acc.replaceFirst(Regex("\\{[^}]*}"), param) }
  }
}
```

Which could then be used like so:
```kotlin
navigationAction.navigateTo(ASSOCIATION.withParams("123"))
```

This would also work with nested routes with different UIDs due to the use of `varargs`.

Zafouche: Wow, nice one, I wouldn't have thought of the use of `varargs`! Thanks for teaching me this. I see you used folding, like we did in Scala!

Zafouche:
I will sync this branch with the `main` branch.

Zafouche:
I ended up simply changing the `String` value of the `ASSOCIATION_PROFILE` like @oskar-codes suggested and added the helper function the `Screen` object.

Redd87 (commented):
Looks good! Though I would add javadoc to the classes and methods  to facilitate their use

Zafouche:
Not sure what else can be tested. @Romainhir will need the changes done to implement the association profile screen further. Does this look good for you to merge? @oskar-codes @Redd87

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
```diff
@@ -10,25 +10,25 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.launch
 
-class ExploreViewModel(val repository: AssociationRepositoryFirestore) : ViewModel() {
+class AssociationViewModel(val repository: AssociationRepository) : ViewModel() {
```
oskar-codes, line 13: Very good call on using the interface as type instead of the Firestore implementation.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
oskar-codes, line 1: Clean code 💪

oskar-codes (approved):
Great work on this! The code is clean, follows best practices, and functions as expected. I’m happy to approve the changes — well done!

PR #63, opened 2024-10-13
Chore/scrape associations

oskar-codes:
This pull request just adds a scraping script to extract associations by category on https://www.epfl.ch/campus/associations/category/.


Zafouche (approved):
Thanks a lot for this @oskar-codes, this is very much needed since this will allow the usage of the repository and the view model on the real data.

PR #32, opened 2024-10-09
Chore/firestore

oskar-codes:
This pull request improves the code concerning Firebase, with the main feature being the `FirestoreReferenceList` class. The class allows Firestore objects to contain UIDs to other objects, that are then fetched when required. This is for example useful for the `Association` class that contains `members: FirestoreReferenceList<User>`.

PR #56, opened 2024-10-12
Feature/strings management

Aurelien9Code:
This PR include strings and language management using standard usage

-> I willingly separated my commits in two main parts : file changes with Resource manager and without 
Because I am not quite sure if creating a Resource Manager is a good pratice. I saw it could be great to prevent passing context to a class containing no `@Composable` functions, but some forums said it could include memory leaks
I will ask the assistants, but i put it here so that anyone can discuss it 

-> The EventListOverview.kt and the Login.kt screens does not include these changes are they are in other pull requests and doing two times the changes would be stupid. 

Thank you !

UPDATE : I will still ask the assistants for my Resource Manager, however i won't link it to the project because it causes big errors that I know how to fix, but I want to be sure that creating a Resource Manager is a good practice and is useful.

PR #68, opened 2024-10-14
Refactor hydration and serialization functions

oskar-codes:
This pull request refactors the existing hydration functions and adds serialization functions to `User`, `Association` and `Event`. This will ensure data is properly formatted when sending/retrieving data from Firestore.

This follows [this comment](https://github.com/SwEnt-Group13/Unio/pull/32#pullrequestreview-2365421530) on #32.

### DIFF
--- a/app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
+++ b/app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
@@ -1,0 +1,45 @@
+package com.android.unio.model.firestore.transform
+
+import com.android.unio.model.association.Association
+import com.android.unio.model.association.AssociationRepositoryFirestore
+import com.android.unio.model.event.Event
+import com.android.unio.model.event.EventRepositoryFirestore
+import com.android.unio.model.user.User
+import com.android.unio.model.user.UserRepositoryFirestore
+
+fun AssociationRepositoryFirestore.Companion.serialize(association: Association): Map<String, Any> {
+  return mapOf(
+      "uid" to association.uid,
+      "url" to association.url,
+      "acronym" to association.acronym,
+      "fullName" to association.fullName,
+      "description" to association.description,
+      "members" to association.members.list.value.map { it.uid })
+}
+
+fun UserRepositoryFirestore.Companion.serialize(user: User): Map<String, Any> {
+  return mapOf(
+      "uid" to user.uid,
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.android.unio.model.user.UserRepositoryFirestore
+
+fun AssociationRepositoryFirestore.Companion.serialize(association: Association): Map<String, Any> {
+  return mapOf(
+      "uid" to association.uid,
+      "url" to association.url,
+      "acronym" to association.acronym,
+      "fullName" to association.fullName,
+      "description" to association.description,
+      "members" to association.members.list.value.map { it.uid })
```
Zafouche, line 17: I know this has been said a lot already, but we'll have to extract these strings into our string value file, as we had discussed in last meeting. This will be required to be done for all the files anyway

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
Zafouche, line 1: Good work on these tests!

Zafouche (approved):
LGTM. Great work again, there is not much to say.

- Strong tests, good job on the excellent code coverage!
- Useful documentation has also been added
- Clean organization

This goes a long way, it brings a much better structure to our project.

app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
Aurelien9Code: Depends if the file contains a `@Composable` function and thus a `Context`. If not, we'll have to wait for the assistant' answer to see if we can create a Resource Manager as explained in #56 : )

Zafouche: You're right, for this specific case I don't think it is even useful as the purpose of this PR is to centralize these things. Adding them on a different file to only use them here is unnecessary

oskar-codes:
The 3 previous commits further enhance this pull request with additions concerning error handling and testing. I will now merge this branch.

oskar-codes:
We really need to find a way to block pushing if the files aren't formatted using ktfmtFormat.

PR #71, opened 2024-10-15
Update EPFL association scraping script

oskar-codes:
This tiny pull request improves the EPFL association scraping script to includes categories directly as a property of the association, as defined in the project. The result of the scraping algorithm has been uploaded to Firebase Firestore.


Romainhir (approved):
LGTM

PR #77, opened 2024-10-16

Romainhir:

- A apk-maker.yml file that will be useful to setup the Workflow


Romainhir:
It seems that a Workflow can only be setup from the main, and I cannot be tested beforehand. That's why I made this PR with this only commit.

Zafouche (approved):
Thanks for this.

PR #78, opened 2024-10-16
Refactor FirestoreReferenceList

oskar-codes:
This pull request adds helper functions and an overall refactor to the `FirestoreReferenceList` class. 
## Helper functions
For example, the helper functions enable you to write:
```kt
Association.emptyFirestoreReferenceList()
```
Instead of the previous:
```kt
FirestoreReferenceList.empty(
      collection = Firebase.firestore.collection(ASSOCIATION_PATH),
      hydrate = AssociationRepositoryFirestore.Companion::hydrate)
```
This will make it easier for others to instanciate a `FirestoreReferenceList`.

## Test refactor
This pull request also changed the type of the `collection` argument of the constructor of `FirestoreReferenceList` to `String` instead of `CollectionReference`. This allows the collection to be resolved at runtime instead of instanciation, which will make testing classes with a `FirestoreReferenceList` property much easier. From now on, tests won't need Robolectric instrumentation and Firebase initialization. This will (and I already see the results locally) greatly speed up the time it takes to run unit tests.

However, an issue emerged from these changes. Because the `collection` argument is now a `String`, there is no way to mock a Firestore database and dependecy inject it. The fix for this is simply to use [MockK](https://mockk.io/) that allows mocking static methods without dependency injection by arguments. That is why this pull request also adds MockK as a dependency.


AlouchLaBouche (changes requested):
In general, great work. This syntax is so much better than before and will definitely save us a lot of time writing boilerplate code. 
Sadly I have to refuse the PR, because you haven't adapted the user data class in the tests. see : `HydrationAndSerializationTest.kt`, `UserRepositoryFirestoreTest.kt` and `UserTest.kt`

Zafouche:
Thank you so much for this 🙏🏽.
Also, don’t really understand what @AlouchLaBouche is talking about either, but maybe he is talking about the merge conflicts? I cannot view them at the moment so I don’t really know what the differences are, just a thought.

AlouchLaBouche:
Yes sorry I was talking about the merge conflicts. I didn't realise you have to merge them after the PR is approved. 
Anyhow, great work 👍,  sorry for the incovenience (better be safe than sorry as we say).

AlouchLaBouche (approved):

PR #80, opened 2024-10-16
test: Write additional tests to increase code coverage for AssociationProfile

Romainhir:
The line coverage of the "AssociationProfile" can be improved with more tests. This is the purpose of this PR. It includes :

  - Test the different button if they do the the correct behavior
  - To do so, we change the placeholder action "Toast.show" with another one that is easier to test. We use instead "SnackBar".

More test will be done when we will link the backend to this screen

### DIFF
--- a/.github/workflows/apk-maker.yml
+++ b/.github/workflows/apk-maker.yml
@@ -1,0 +1,40 @@
+name: APK Maker
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+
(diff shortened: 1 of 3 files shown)


Zafouche (approved):
LGTM. I'll have to find out more about the snackbar with some research, but interesting!
Also, is it normal that these lines, which contain test tags, aren't tested?

Screenshot 2024-10-17 at 10 28 13

PR #84, opened 2024-10-17

Romainhir:

    A apk-maker.yml file that will be useful to setup the Workflow

The apk-maker already exist on the main branck but need to be fix. That's differenciate this PR from the precedent one


Redd87 (approved):
LGTM

PR #85, opened 2024-10-17
Feature: account creation screen

AlouchLaBouche:
This feature aims to add the account creation screen as seen on the [Figma](https://www.figma.com/design/rlvqRyIsbdSyEbXu8Xp7yG/Unio?node-id=1-2&node-type=canvas&t=yFwKcVquSXymBAye-0). It is now possible to add more details about the user such as first name, last name, a biography, and a list of interests when first logging in to our app. The UI is linked to the backend, so the information is stored in FireStore!  

This PR includes:
- `AccountDetails.kt` Screen
- `OverlayCard.kt` Screen (corresponding to the dialog seen in figma)
- UI tests
- Fixes, such as setting dynamic color to false in the app theme (this ruins our chosen color palette)
- A new update user function made with the serialise method (thanks @oskar-codes )  
- Updated navigation accordingly

What is still to do: 
- Being able to add a list of socials (Instagram, Website, ect...)
- Being able to add a profile picture (from you gallery or directly by taking a selfie)

Potential problems: 

When first logging in, there a few checks that are missing.
Firstly it would be good to check that the `Firebase.auth.currentUser` is not null when first creating the account (so that we are sure that the email verification went well). Sadly I didn't manage to mock the `Firebase.auth.currentUser` for the tests, that's why the field is left commented (and also why one of my tests is commented).
Secondly, the navigation is pretty straight foward : `Welcome -> AccountDetails -> Home`, but if there user gets interrupted in this process and comes back later, he will go straight to the `Home` page whitout finishing adding his account details (as we only check that his email is verified to go to `AccountDetails`)


oskar-codes:
Great work on this Alexei! About mocking `Firebase.auth.currentUser`, I've rencently merged into main a new dependency you could use, namely [MockK](https://mockk.io/). It allows you to mock static properties (like `Firebase.auth`), without having to inject the mock by parameter. You might want to look into that.

Also it appears that a test fails in the CI, but when testing on my machine the test passes which is weird. I will gladly review your request once you find a way to fix that.

AlouchLaBouche:
Thanks for the feedback @oskar-codes, and sorry for the late response. I managed to mock `Firebase.auth.currentUser` and now all the tests pass. The problem resided in the fact that I tried to mock an abstract method `.getUid()`, but that is impossible, so I had to mock it's implementation which was found in a class called `zacc` which extends `FirebaseUser`.
On top of that I made the methods from `navigationAction` open, exactly as we did in the [Bootcamp](https://github.com/swent-epfl/public).

app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -48,14 +50,16 @@ fun UnioApp() {
...
+
   val context = LocalContext.current
 
   // Redirect user based on authentication state
   Firebase.auth.addAuthStateListener { auth ->
     val user = auth.currentUser
     if (user != null) {
       if (user.isEmailVerified) {
-        navController.navigate(Route.HOME)
+        navController.navigate(Screen.ACCOUNT_DETAILS)
```
oskar-codes, line 62: Later on, we will need to implement some logic that checks if the account details have already been given. This will ensure users that exit and relaunch the app during account details creation will still need to complete them before accessing the application.

app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -68,7 +72,9 @@ fun UnioApp() {
     navigation(startDestination = Screen.WELCOME, route = Route.AUTH) {
       composable(Screen.WELCOME) { WelcomeScreen(navigationActions) }
       composable(Screen.EMAIL_VERIFICATION) { EmailVerificationScreen(navigationActions) }
-      composable(Screen.ACCOUNT_DETAILS) { AccountDetails(navigationActions) }
+      composable(Screen.ACCOUNT_DETAILS) {
+        AccountDetails(navigationActions, userRepositoryFirestore)
```
oskar-codes, line 76: Later on, this might need to be changed to a `UserViewModel`. I'm currently working on it and I'll adapt this as soon as possible.

app/src/main/java/com/android/unio/ui/authentication/OverlayCard.kt
```diff
@@ -0,0 +1,113 @@
...
+import androidx.compose.ui.window.DialogProperties
+import com.android.unio.model.user.Interest
+import com.android.unio.ui.theme.AppTypography
+import kotlinx.coroutines.flow.MutableStateFlow
+
+@Composable
+fun InterestOverlay(
+    onDismiss: () -> Unit,
+    onSave: () -> Unit,
+    interests: MutableStateFlow<MutableList<Pair<Interest, MutableState<Boolean>>>>
```
oskar-codes, line 39: To reduce cognite complexity, parameter type could be changed to a `typealias`. This will ensure better readability.

app/src/main/java/com/android/unio/ui/theme/Theme.kt
```diff
@@ -266,7 +266,7 @@ val unspecified_scheme =
 fun AppTheme(
     darkTheme: Boolean = isSystemInDarkTheme(),
     // Dynamic color is available on Android 12+
-    dynamicColor: Boolean = true,
+    dynamicColor: Boolean = false,
```
oskar-codes, line 269: Great work on finding the fix to this issue 💪

oskar-codes (approved):
The code is generally compliant with best practices and goes in the right direction towards the development of Unio.

PR #86, opened 2024-10-17

Romainhir:

    A apk-maker.yml file that will be useful to setup the Workflow

The "$GOOGLE_SERVICES" and the "$LOCAL_PROPERTIES" were not loaded. This is why the CI crashed when building the apk. This PR is here to fix this problem

### DIFF
--- a/.github/workflows/apk-maker.yml
+++ b/.github/workflows/apk-maker.yml
@@ -35,6 +35,19 @@ jobs:
           restore-keys: |
             gradle-${{ runner.os }}-
 
+      # Load google-services.json and local.properties from the secrets
+      - name: Decode secrets
+        env:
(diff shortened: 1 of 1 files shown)


oskar-codes (approved):
Nice addition, creating APK files will be very useful. By the way, have you tested an APK file on your phone to see if it works properly ? I'll approve the PR but it's always good to do some real-circumstances testing.

Romainhir:
> Nice addition, creating APK files will be very useful. By the way, have you tested an APK file on your phone to see if it works properly ? I'll approve the PR but it's always good to do some real-circumstances testing.

I did not manually install it on my phone, but instead of using an emulator to make tests when I'm coding, I use directly my phone.

PR #90, opened 2024-10-19
Fix color theme for Bottom Nav Menu

AlouchLaBouche:
Very small PR that aims to solve the color theme of the Bottom Nav Menu. This simply includes switching the composable functions of `material` to the equivalent but of `material3` such that the composables are in accordance with our `AppTheme`.


Aurelien9Code (approved):
Looks perfect ! 
Now the bottom bar will look a bit better when we'll work ; )

PR #95, opened 2024-10-21
Refactor/user data class

AlouchLaBouche:
This PR includes the refactored user data class, new navigation logic and all the refactoring needed for the tests. 
For the user data class I added 2 new fields, those being : `joinedAssociations` and `hasProvidedAccountDetails`, for the following reasons: 
- It is now possible to see what associations a user is part of.
- The latter is to ensure correct navigation logic. If in the user has not completed giving his account details but has verified his email, he will be redirected to the `AccountDetails` Screen instead of straight to `Home`.

Potential Risks:
- The `joinedAssociations` Reference List might be desynced with the `members` Reference List found in the association data class. (We partly loose the [SSOT](https://en.wikipedia.org/wiki/Single_source_of_truth) concept).

### DIFF
--- a/app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
+++ b/app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
@@ -37,13 +37,17 @@ fun UserRepositoryFirestore.Companion.hydrate(data: Map<String, Any>?): User {
   val followingAssociationsUids = data?.get("followingAssociations") as? List<String> ?: emptyList()
   val followingAssociations = Association.firestoreReferenceListWith(followingAssociationsUids)
 
+  val joinedAssociationsUids = data?.get("joinedAssociations") as? List<String> ?: emptyList()
+  val joinedAssociations = Association.firestoreReferenceListWith(joinedAssociationsUids)
+
   return User(
       uid = data?.get("uid") as? String ?: "",
       email = data?.get("email") as? String ?: "",
       firstName = data?.get("firstName") as? String ?: "",
       lastName = data?.get("lastName") as? String ?: "",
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
oskar-codes, line 1: Good error handling, well done! 💪

app/src/main/java/com/android/unio/model/user/User.kt
```diff
@@ -35,21 +35,24 @@ data class UserSocial(val social: Social, val content: String)
  * @param firstName The first name of the user.
  * @param lastName The last name of the user.
  * @param biography The biography of the user.
- * @param followingAssociations The associations that the user is following.
+ * @param followedAssociations The associations that the user is following.
  * @param interests The interests of the user.
```
oskar-codes, line 39: You forgot to add `joinedAssociations` to the documentation

oskar-codes (changes requested):
This pull request is generally good, however there are missing features that render it unfit for merge.

app/src/main/java/com/android/unio/model/user/User.kt
```diff
@@ -36,6 +36,7 @@ data class UserSocial(val social: Social, val content: String)
  * @param lastName The last name of the user.
  * @param biography The biography of the user.
  * @param followedAssociations The associations that the user is following.
+ * @param joinedAssociations The associations that the user is member of.
```
oskar-codes, line 39: Nice addition! This will ensure scalability and maintainability.

oskar-codes (approved):
Nice refactor! This will be very useful later on.

PR #98, opened 2024-10-21
Add quality of life features to authentication system

oskar-codes:
This pull request adds several minor features that enhance the user experience when signing up or logging into Unio. These include:
- Trimming whitespace around the email and password.
- Adding password requirements and clearly showing them to the user.
- Configuring email keyboard and password keyboard for each field (disabling autocomplete etc.)
- Check for internet connectivity before making Firebase requests and notify the user.

These changes are minor and should be able to be merged with no issue.


AlouchLaBouche (approved):
Nice quality of life features 👍 . Good job. Just watch out when trimming the password... People might want to have withespaces in it? In general, I think it's a bad idea to modify the passwords without the user knowing.

PR #99, opened 2024-10-23
Feature/strings management v2

Aurelien9Code:
This new version of the string management kills the Resource Manager, which was in the end a bad idea. 
As the assistant adviced, I created a Kotlin class to handle all harcoded strings that can not be into strings.xml (so are not part of a `@Composable` class) as constants.   
I did NOT put logs, mock classes hardcoded strings (as they might be deleted one day), or parameter names in the hardcoded strings structure. 
I hope it's good for you ! Then we will all use this to a more readable code.


AlouchLaBouche:
Hey, thanks for your work, this will be useful for all UI elements of our app and should quickly be incoroporated to handle all the different strings in different languages of our app. But I need to ask, why add a kotlin class with all the hardcoded strings instead of using the already existing `.xml` files ? You say that it was a bad idea in the end to use the `.xml` files but I just switched to your branch and tried to incoporate the string form the `.xml` files with `context.getString(R.string.app_name)` for example, and the app worked just fine. 
Using the `.xml` files would be especially useful as it will be very easy to change the language of the application as the strings will be automatically found.

Aurelien9Code:
> Hey, thanks for your work, this will be useful for all UI elements of our app and should quickly be incoroporated to handle all the different strings in different languages of our app. But I need to ask, why add a kotlin class with all the hardcoded strings instead of using the already existing `.xml` files ? You say that it was a bad idea in the end to use the `.xml` files but I just switched to your branch and tried to incoporate the string form the `.xml` files with `context.getString(R.string.app_name)` for example, and the app worked just fine. Using the `.xml` files would be especially useful as it will be very easy to change the language of the application as the strings will be automatically found.

Hello, and thanks for your review! I apologize for the confusion—I thought I had clarified this in our previous meeting, but I realize now I wasn’t clear enough. Let me explain in more detail:

The strings.xml file is still available in both languages and can be accessed using `context.getString(R.string.app_name)` (as you correctly pointed out!). However, not all classes have access to `Context`. Only classes containing `@Composable `functions, typically UI classes, can have a `Context`.

In the first version of this PR, I had a similar thought to yours and created a `Resource Manager` to provide `Context` to every `non-@Composable` class. However, this approach can lead to memory leaks, so it’s not ideal.

This makes sense when we consider that users don’t interact with the hardcoded strings in 'functional/backend' classes, nor do they need these to be language-dependent. To avoid hardcoded strings in these functional classes for our internal purposes, I created a dedicated class to handle them as constants.

After researching this further and considering the assistant’s advice from our last meeting, I’m confident that separating UI strings in strings.xml and placing backend strings in a constants file is the right approach. If you’d like to discuss it further, we can bring it up in the team meeting on Tuesday!

AlouchLaBouche:
Oh so if I understood correctly, the `.xml` files will be used for strings that are displayed in the UI (typically all the strings found in `@Composable` functions), while the `Strings.kt` file will be used for all strings related to the backend of our poject such as URL paths or other definitions which will not be visible to the users when using the app ?

Aurelien9Code:
> Oh so if I understood correctly, the `.xml` files will be used for strings that are displayed in the UI (typically all the strings found in `@Composable` functions), while the `Strings.kt` file will be used for all strings related to the backend of our poject such as URL paths or other definitions which will not be visible to the users when using the app ?

Yes, you are completely right ! ; ) It's the best way to structure the strings management !

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -280,23 +277,25 @@ fun AssociationProfileEvents(context: Context) {
...
               Spacer(Modifier.width(2.dp))
-              Text(getString(R.string.association_see_more))
+              Text(context.getString(R.string.association_see_more))
             }
       }
 }
 
 @Composable
 fun AssociationEventTitle() {
+  val context = LocalContext.current
```
Romainhir, line 287: This val could be passed as an argument to avoid duplicate code (already in the AssociationProfileScaffold)

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -280,23 +277,25 @@ fun AssociationProfileEvents(context: Context) {
...
   Text(
-      getString(R.string.association_upcoming_events),
+      context.getString(R.string.association_upcoming_events),
       modifier = Modifier.padding(horizontal = 20.dp).testTag("AssociationEventTitle"),
       style = AppTypography.headlineMedium)
 }
 
 @Composable
 fun AssociationDescription() {
+  val context = LocalContext.current
```
Romainhir, line 296: Same, already in the AssociationProfileScaffold

Romainhir (approved):
Small comment on context in AssociationProfile but otherwise LGTM

PR #104, opened 2024-10-28
Link the association profile UI to the backend

Romainhir:
Actually, the association profile UI only has temporary data that is the same for any association we click on. This task aims to solve this by linking the UI to the view model/repository. These classes are already implemented, the screen just needs to feed the data to the correct component of the UI. A good example is given in the Figma. This include :

-     The title of the association
-     The number of followers/members
-     The association description
-     The upcoming events
-     The contact members


app/src/main/java/com/android/unio/model/association/Association.kt
```diff
@@ -21,24 +22,25 @@ data class Association(
...
+  EPFL_STUDENTS(AssociationStrings.EPFL_STUDENTS),
+  COUNTRIES(AssociationStrings.COUNTRIES),
+  SUSTAINABILITY(AssociationStrings.SUSTAINABILITY),
+  SCIENCE_TECH(AssociationStrings.SCIENCE_TECH),
+  CULTURE_SOCIETY(AssociationStrings.CULTURE_SOCIETY),
+  ARTS(AssociationStrings.ARTS),
+  ENTERTAINMENT(AssociationStrings.ENTERTAINMENT),
+  SPORTS(AssociationStrings.SPORTS),
+  GUIDANCE(AssociationStrings.GUIDANCE),
+  UNKNOWN(AssociationStrings.UNKNOWN)
```
armouldr, line 45: It's nice to get rid of these harcoded strings :)

app/src/main/java/com/android/unio/model/event/EventType.kt
```diff
@@ -11,17 +12,16 @@ import com.android.unio.ui.theme.eventTypeSport
...
-  TRIP(eventTypeTrip, "trip"),
-  OTHER(eventTypeOther, "other"); // Default color
+  FESTIVAL(eventTypeFestival, EventTypeStrings.FESTIVAL),
+  APERITIF(eventTypeAperitif, EventTypeStrings.APERITIF),
+  NIGHT_PARTY(eventTypeNightParty, EventTypeStrings.NIGHT_PARTY),
+  JAM(eventTypeJam, EventTypeStrings.JAM),
+  NETWORKING(eventTypeNetworking, EventTypeStrings.NETWORKING),
+  SPORT(eventTypeSport, EventTypeStrings.SPORT),
+  TRIP(eventTypeTrip, EventTypeStrings.TRIP),
+  OTHER(eventTypeOther, EventTypeStrings.OTHER); // default
```
armouldr, line 22: Same, it's great to have no more hardcoded values !

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -137,14 +131,22 @@ fun AssociationProfileScaffold(
       },
       topBar = {
         TopAppBar(
-            title = { Text(title, modifier = Modifier.testTag("AssociationProfileTitle")) },
+            title = {
+              val title: String
+              if (association != null) {
+                title = association.fullName
```
armouldr, line 137: As I said in above, I think that it would look better and more coherent if we used association.name here instead.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -160,39 +162,65 @@ fun AssociationProfileScaffold(
...
+ *
+ * @param association (Association) : The association currently displayed
+ * @param context (Context) : The context of the screen
+ */
 @Composable
-fun AssociationRecruitment(context: Context) {
+private fun AssociationRecruitment(association: Association, context: Context) {
   Text(
-      text = getString(R.string.association_join) + " <Association> ?",
+      text = context.getString(R.string.association_join) + " ${association.fullName} ?",
```
armouldr, line 218: same here, the association.fullName isn't always coherent

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -225,98 +253,165 @@ fun AssociationRecruitment(context: Context) {
...
-      Icon(Icons.Filled.Person, contentDescription = "user's profile picture", Modifier.size(36.dp))
-      Text(text = "<Casey Rue>", style = AppTypography.headlineSmall)
+      Row(
+          horizontalArrangement = Arrangement.spacedBy(115.dp, Alignment.Start),
+          verticalAlignment = Alignment.CenterVertically,
+      ) {
+        Icon(
+            Icons.Filled.Person,
+            contentDescription = "user's profile picture",
+            Modifier.size(36.dp))
```
armouldr, line 292: User has a profilePicture value, why not directly implement it ? maybe it could be a small separate task, but then I fell like we should add it to the scrum board for this sprint

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -225,98 +253,165 @@ fun AssociationRecruitment(context: Context) {
...
+    association: Association,
+    associationViewModel: AssociationViewModel,
+    context: Context
+) {
+  var isSeeMoreClicked by remember { mutableStateOf(false) }
+  var events = emptyList<Event>()
+  associationViewModel.getEventsForAssociation(association) { fetchedEvents ->
+    events = fetchedEvents
+  }
+  if (events.isEmpty()) {
```
armouldr, line 318: I couldn't test because as of now no association has any events, but I think that events.isEmpty() will always return true because it does not wait for getEventsForAssociation(association) to finish. Tell me if I am wrong but in my understanding the condition and following code should be in the onSuccess lambda of getEventsForAssociation

armouldr (changes requested):
Overall a great add to the project, with good and readable code and useful javadoc :)
However, I spotted a few things to discuss:
- While running the app, I saw that some events don't have their name as a header but instead have their category (for ex Artiphys and Balelec have "Festival" as a title, which doesnt seem right. Also 180°C becomes "Association to promote cooking amongst cooking". More generally I think that we shoudn't use the associations full name attribute as the main header but instead use their short version, and maybe add the long version as a shorttext under it ?

-As detailed in a comment below, I don't think that the events will be displayed when there are some, due to the events.isEmpty verification no being in onSuccess lambda function. 

-The user profile picture is not displayed, but maybe it can be considered as a different task that's for you to say :)

Otherwise great work!

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
Romainhir: Maybe this will never happend, but in case this will, the app will not crash and it will just display some text. That's why I added this. Tell me if it should be deleted even if we consider this

app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -51,7 +51,8 @@ fun UnioApp() {
   val db = FirebaseFirestore.getInstance()
 
   val associationRepository = AssociationRepositoryFirestore(Firebase.firestore)
-  val associationViewModel = AssociationViewModel(associationRepository)
+  val eventRepository = EventRepositoryFirestore(Firebase.firestore)
```
Redd87, line 54: This should be refactored to use the db defined above, but I've fixed this in my PR so it should be fixed by merging

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
```diff
@@ -23,22 +29,40 @@ class AssociationViewModel(val repository: AssociationRepository) : ViewModel()
...
+                AssociationRepositoryFirestore(Firebase.firestore),
+                EventRepositoryFirestore(Firebase.firestore))
+                as T
           }
         }
   }
 
+  fun getEventsForAssociation(association: Association, onSuccess: (List<Event>) -> Unit) {
+    viewModelScope.launch {
+      Log.d("ExploreViewModel", "Getting events for association ${association.fullName} !!!")
```
Redd87, line 50: This log probably should be removed before production?

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
```diff
@@ -23,22 +29,40 @@ class AssociationViewModel(val repository: AssociationRepository) : ViewModel()
...
-            return AssociationViewModel(AssociationRepositoryFirestore(Firebase.firestore)) as T
+            return AssociationViewModel(
+                AssociationRepositoryFirestore(Firebase.firestore),
+                EventRepositoryFirestore(Firebase.firestore))
+                as T
           }
         }
   }
 
+  fun getEventsForAssociation(association: Association, onSuccess: (List<Event>) -> Unit) {
```
Redd87, line 48: Event though it is pretty obvious what is does by the name, this function (and others) should still be commented with a bit of javadoc

Redd87 (approved):
The PR looks good to me, most of the problems I found were really minor

armouldr (approved):
Thanks for the changes @Romainhir LGTM :)

PR #105, opened 2024-10-28
chore: implement caching guidelines for images

armouldr:
Implemented caching for images (the default options were used before) and fixed a bug that made the app get events multiple times.

The bug in question was declaring our EventListViewModel inside a composable f the app's Navhost, which is internally called multiple times.


oskar-codes (approved):
Not much to say about this, overall solid code that adheres to best practices and will improve Unio's performance.

oskar-codes:
Make sure to properly merge `MainActivity`, I believe you should only need the version on main and can discard the one on this branch.

PR #109, opened 2024-10-29
Refactor User data class - Remove pointless attribute 

AlouchLaBouche:
Small PR that refactors the user data class as we've noticed the the `hasProvidedAccountDetails` boolean attribute is completely redundant. We can instead use any other required field such as `FirstName` or `LastName` to check that the user has indeed filled out his `AccountDetails`. This ensures that the navigation still works properly.

This PR includes : 
- Changed User data class.
- All the subsequent modifications for tests and other methods.

### DIFF
--- a/app/src/main/java/com/android/unio/MainActivity.kt
+++ b/app/src/main/java/com/android/unio/MainActivity.kt
@@ -66,7 +66,7 @@ fun UnioApp() {
         userRepositoryFirestore.getUserWithId(
             user.uid,
             {
-              if (it.hasProvidedAccountDetails) {
+              if (it.firstName.isNotEmpty()) {
                 navigationActions.navigateTo(Screen.HOME)
               } else {
                 navigationActions.navigateTo(Screen.ACCOUNT_DETAILS)
--- a/app/src/androidTest/java/com/android/unio/ui/user/UserProfileTest.kt
+++ b/app/src/androidTest/java/com/android/unio/ui/user/UserProfileTest.kt
@@ -54,8 +54,7 @@ class UserProfileTest {
               listOf(
                   UserSocial(Social.INSTAGRAM, "Instagram"),
                   UserSocial(Social.WEBSITE, "example.com")),
-          profilePicture = "https://www.example.com/image",
-          hasProvidedAccountDetails = true)
+          profilePicture = "https://www.example.com/image")
 
   @Before
   fun setUp() {
--- a/app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
+++ b/app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
@@ -54,8 +54,7 @@ fun UserRepositoryFirestore.Companion.hydrate(data: Map<String, Any>?): User {
           (data?.get("socials") as? List<Map<String, String>> ?: emptyList()).map {
             UserSocial(Social.valueOf(it["social"] ?: ""), it["content"] ?: "")
           },
-      profilePicture = data?.get("profilePicture") as? String ?: "",
-      hasProvidedAccountDetails = data?.get("hasProvidedAccountDetails") as? Boolean ?: false)
+      profilePicture = data?.get("profilePicture") as? String ?: "")
 }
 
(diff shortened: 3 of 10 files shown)


Aurelien9Code:
Very good idea ! That makes so much sense now. 🚀

However, I would advise the creation of a `fun hasProvidedAccountDetails` that checks that every mandatory elements are completed, not just the firstName. Normally, firstName would be sufficent, but if 
1) There's any problem or deletion in the database for a User element
2) We add a new mandatory element to the User, we can just add the element to the freshly made `fun hasProvidedAccountDetails` and it will automatically ask every user that already exists but don't have the require information yet to fill it (as it will put them on the AccountDetails screen when going back to the app)

Aurelien9Code (commented):
Very good idea ! That makes so much sense now. 🚀

However, I would advise the creation of a fun hasProvidedAccountDetails that checks that every mandatory elements are completed, not just the firstName. Normally, firstName would be sufficent, but if

- There's any problem or deletion in the database for a User element
- We add a new mandatory element to the User, we can just add the element to the freshly made fun hasProvidedAccountDetails and it will automatically ask every user that already exists but don't have the require information yet to fill it (as it will put them on the AccountDetails screen when going back to the app)

AlouchLaBouche:
Thanks for your review @Aurelien9Code. If I understand correctly, you want to add a method that checks if the user has provided his account details. This method arleady exists and is called `checkNewUser` in `User.kt`. For now, this method checks if the first and last name fields are empty and adds the corresponding error tags. 
This method is indeed not perfect, especially because the check is only done client side (and as we know in computer security, never trust the client), but it still restricts a minimum the creation of an Account. You can Check it out for yourself, you are unable to create an account whitout completing these fields. Thus we ensure that the user has a first and last name field.

Aurelien9Code:
> Thanks for your review @Aurelien9Code. If I understand correctly, you want to add a method that checks if the user has provided his account details. This method arleady exists and is called `checkNewUser` in `User.kt`. For now, this method checks if the first and last name fields are empty and adds the corresponding error tags. This method is indeed not perfect, especially because the check is only done client side (and as we know in computer security, never trust the client), but it still restricts a minimum the creation of an Account. You can Check it out for yourself, you are unable to create an account whitout completing these fields. Thus we ensure that the user has a first and last name field.

Okay got it ! Then I would advise to modify you check in MainActivity.kt with if(checkNewUser...) instead of only checking for the firstName for the same twos reasons as above

Aurelien9Code (approved):
Please modify MainActivity.kt before merging : )

app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -66,7 +66,7 @@ fun UnioApp() {
         userRepositoryFirestore.getUserWithId(
             user.uid,
             {
-              if (it.hasProvidedAccountDetails) {
+              if (it.firstName.isNotEmpty()) {
```
oskar-codes, line 69: As @Aurelien9Code mentionned, checking that the `firstName` is not empty isn't entirely foolproof, or even semantically correct. Conceptually, the check should be on the existence of the document at `users/${user.uid}`. As mentionned in [the documentation](https://firebase.google.com/docs/firestore/query-data/get-data#kotlin+ktx_2) this is done inside the `.addOnSuccessListener` with a check to `document.exists()`. 

However, seeing the docs, I believe we should refactor all Firestore repositories to handle the case where the document does not exist. As it currently is implemented, all the success callbacks are called even when a document doesn't exist, but the non-existence of a document should be considered a failure.

oskar-codes (approved):
I will approve this pull request as the code is working (if we do not consider security flaws), but we strongly need to consider refactoring all Firestore repositories to handle the case where a document does not exist.

app/src/main/java/com/android/unio/MainActivity.kt
AlouchLaBouche: You're Right, checking for the existence of the first name is on top of being semantically incorrect, a very bad practise. We should work on that for the next Sprint. Typically a method in the User repo that checks the existence of the document will be a good idea as you explained.

PR #117, opened 2024-10-30
Feat/search : implement the search backend for events and associations

Redd87:
This PR implements the search backend of the project using Jetpack's AppSearch. 

There is a local search database in the SearchRepository which is created on startup based on the data from the Association- and EventRepository, and that database can add, remove documents, and most importantly be queried. The ranking strategy - the order in which matching results will show -  is currently set to none; as long as there is a matching keyword the result will show. This can later be modified to promote or hide certain documents. 
In the Event and Association classes, I created an annex object EventDocument and AssociationDocument, which consists of the attributes on which to search.
The SearchViewModel exposes the query function of the SearchRepository, and allows access to the result data in its StateFlow. 
The SearchRepository is accompanied with a SearchRepositoryTest file containing Unit tests, but the SearchViewModel is not. It will need to have integrations test as unit tests do not make sense in view of the file. This will be done at a later time as this PR is already quite large.
The main has been modified to properly set up the search backend, and is currently injected into the explore page.


app/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
     alias(libs.plugins.gms)
     alias(libs.plugins.sonar)
     id("jacoco")
+    kotlin("kapt")
```
Zafouche, line 13: j'ai pas capté 🤣🤣🤣🤣🤣🤣

app/src/main/java/com/android/unio/model/search/SearchRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Adds the given associations to the search database.
+   *
+   * @param associations the list of [Association] to add
+   */
+  // Should be private, but I need it public for tests
+  fun addAssociations(associations: List<Association>) {
+    Log.d("SearchRepository", "Adding associations to search database")
```
Zafouche, line 92: Don't forget to remove this debug `Log`!

app/src/main/java/com/android/unio/model/search/SearchRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Adds the given events to the search database.
+   *
+   * @param events the list of [Event] to add
+   */
+  // Should be private, but I need it public for tests
+  fun addEvents(events: List<Event>) {
+    Log.d("SearchRepository", "Adding events to search database")
```
Zafouche, line 108: Another debug `Log`!

app/src/main/java/com/android/unio/model/search/SearchViewModel.kt
```diff
@@ -0,0 +1,91 @@
...
+
+  private val _events = MutableStateFlow<List<Event>>(emptyList())
+  val events: StateFlow<List<Event>> = _events
+
+  /** Initializes the ViewModel by creating the search database and connecting it to the session. */
+  init {
+    Firebase.auth.addAuthStateListener {
+      if (it.currentUser != null) {
+        viewModelScope.launch { repository.init() }
+      }
```
Zafouche, line 34: Have a look at how the `init` block is defined in `AssociationViewModel`:
```kotlin
init {
    associationRepository.init { getAssociations() }
  }
``` 

It simply calls the repository's `init` function. You'll see that the addition of the Auth State Listener is done in that repository `init` function, e.g. in `AssociationRepositoryFirestore`:
```kotlin
override fun init(onSuccess: () -> Unit) {
    Firebase.auth.addAuthStateListener {
      if (it.currentUser != null) {
        onSuccess()
      }
    }
  }
``` 

I think it makes more sense to add the Auth State Listener there as it is the database's duty, but let me know if you disagree

app/src/test/java/com/android/unio/model/search/SearchRepositoryTest.kt
Zafouche, line 1: Strong tests! 💪

gradle/libs.versions.toml
```diff
@@ -17,7 +20,7 @@ lifecycleLivedataKtx = "2.8.6"
 materialIconsExtended = "1.7.3"
 mockitoAndroid = "5.13.0"
 mockkMockk = "1.9.3"
-mockkVersion = "1.13.0"
+mockkVersion = "1.13.7"
```
Zafouche, line 23: Good call on upgrading the version, I see nothing breaks!

Zafouche (commented):
Thank you for your amazing work on this difficult task! 🚀

Overall, great code! Good start on the tests as well, since I know you plan on adding more in the next Sprint, no need to comment on the code coverage.

The only thing that bothers me is the choice of architecture for the Search view model and repository. In brief, I think it violates the MVVM pattern, because in my opinion the Search repository shouldn't be able to utilize the `AssociationRepository` and the `EventRepository`. I think what would be a correct pattern would be to have the Search view model utilize the `AssociationViewModel` and the `EventViewModel`. This would require passing the `onSuccess` and `onFailure` callbacks through the function parameters to the Search repository, so we would have the classic `ViewModel`-`Repository` relationship we have in all other such pairs.
Have a look at my comments on the respective files. Please explain to me your choice in those, and I will gladly approve this PR!
 
Other than that, you'll find my comments to be on small details which aren't very important. However don't forget to remove the debugging `Log.d` calls, and optimize the imports!

app/src/main/java/com/android/unio/model/search/SearchRepository.kt
```diff
@@ -0,0 +1,211 @@
...
+import kotlinx.coroutines.withContext
+
+/** Repository for searching associations and events */
+class SearchRepository(
+    private val appContext: Context,
+    private val associationRepository: AssociationRepository,
+    private val eventRepository: EventRepository
+) {
+  // Should be private, but I need it public for tests
+  var session: AppSearchSession? = null
```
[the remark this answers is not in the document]
oskar-codes: Tests should focus only on the input/output of classes, **not** the internal behavior. This means that either this property should be private and not tested, or public and part of the SearchRepository's API. Testing internal behavior is bad practice as it might be subject to change at any moment.

Redd87:
> Thank you for your amazing work on this difficult task! 🚀
> 
> Overall, great code! Good start on the tests as well, since I know you plan on adding more in the next Sprint, no need to comment on the code coverage.
> 
> The only thing that bothers me is the choice of architecture for the Search view model and repository. In brief, I think it violates the MVVM pattern, because in my opinion the Search repository shouldn't be able to utilize the `AssociationRepository` and the `EventRepository`. I think what would be a correct pattern would be to have the Search view model utilize the `AssociationViewModel` and the `EventViewModel`. This would require passing the `onSuccess` and `onFailure` callbacks through the function parameters to the Search repository, so we would have the classic `ViewModel`-`Repository` relationship we have in all other such pairs. Have a look at my comments on the respective files. Please explain to me your choice in those, and I will gladly approve this PR!
> 
> Other than that, you'll find my comments to be on small details which aren't very important. However don't forget to remove the debugging `Log.d` calls, and optimize the imports!

Thanks for the feedback, I'll do these changes ASAP. The justification behind the SearchRepository is that it acts as a composite repository for the Association and EventRepository; however I honestly simply didn't consider doing it as you suggested. I don't feel that the SearchRepository violates the MVVM model, as it receives the other repositories by dependency injection and simply aggregates their data, there is no interaction between the repositories, but I do agree that the SearchViewModel is quite different to the other ViewModels, but I don't know if that itself is a violation of the MVVM. I could refactor it to adhere to the others but that would take quite a lot of time and I'm not sure if it's that advantageous

Let me know if you have other questions

Romainhir (approved):

app/src/main/java/com/android/unio/model/association/Association.kt
```diff
@@ -43,3 +48,27 @@ enum class AssociationCategory(val displayName: String) {
...
+    @Id val uid: String,
+    @StringProperty(indexingType = StringPropertyConfig.INDEXING_TYPE_PREFIXES)
+    val url: String = "",
+    @StringProperty(indexingType = StringPropertyConfig.INDEXING_TYPE_PREFIXES)
+    val name: String = "",
+    @StringProperty(indexingType = StringPropertyConfig.INDEXING_TYPE_PREFIXES)
+    val fullName: String = "",
+    @StringProperty(indexingType = StringPropertyConfig.INDEXING_TYPE_PREFIXES)
+    val description: String = ""
+    // TODO add members
```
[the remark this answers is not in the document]
Redd87: Yes these are all searchable, and I disagree, as I feel like you might not sometimes remember the association name but want can then find it with keywords that are in the description

PR #119, opened 2024-10-30
Fix Navigation Bug on Device Rotation

Zafouche:
## Description

It was noted in the MS1 grading that there was an issue with device orientation: When the device is rotated in the Explore page, the app is redirected to the Home Screen.
The same bug is present when we switch from light to dark mode, and vice-versa.

Testing this shows that the bug is present in all screens. The bug cannot only be in the Explore screen as it does not define the behavior of our `NavHost`.
The addition of the auth state listener to is the problem. It is done each time the `UnioApp` composable is recomputed. 

### Edit 1:
Extracting the logic from the MainActivity and putting it into WelcomeScreen with the rest of the login logic fixes the bug.
I am still having issues with writing tests for this logic, but the bug is gone!
⚠️ This modification also fixes the dark mode activation bug, which also caused the app to take us back to the Home Screen. It is expected that it fixes all bugs of this type, e.g. due to the refresh.

### Edit 2:
The changes from `Edit 1` were not compatible with new updates that were made in the main branch during the development of this bug fix. 
This is what I ended up doing: 
- Created an the view model `AuthViewModel`, that handles the logic of the Auth State listener and its state
- Integrated the view model in the `MainActivity` in a manner that fixes the bug, e.g. it remembers the state of the login and which screen it was on. 
This solution works and the view model has been tested.

## Testing
The login logic in `AuthViewModel` has been tested to ensure the state transitioning is correct.


Zafouche:
I am struggling to mock the `FirebaseUser` and its properties, e.g. mocking this (as it is an abstract method, classic mockk does not suffice, and I am still researching how to mock it): 
```kotlin
firebaseUser.isEmailVerified
```

Zafouche:
Ready to be reviewed and merged.

Redd87 (approved):
A solid pull request, good work overall! When you merge you'll have conflict in the Main as I reordered the initialization of the repos and viewmodels, so be aware of that.

Zafouche:
Fix does not work after updating the branch with main. Can't quite figure out why.

Zafouche:
A solution to the bug was found through the usage of an authentication view model. Please see `Edit 2` of the main description of the PR quoted below:

> ### Edit 2:
> The changes from `Edit 1` were not compatible with new updates that were made in the main branch during the development of this bug fix. This is what I ended up doing:
> 
> * Created an the view model `AuthViewModel`, that handles the logic of the Auth State listener and its state
> * Integrated the view model in the `MainActivity` in a manner that fixes the bug, e.g. it remembers the state of the login and which screen it was on.
>   This solution works and the view model has been tested.

This effectively fixes the bug that redirects the user to the Home Screen whenever the app composable was refreshed, caused by, for example, a switch between light and dark mode or a rotation of the screen.

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+
+  init {
+    verifyAuthState()
+  }
+
+  /**
+   * Verifies the authentication state of the user. If the user should be redirected, the
+   * [authState] is updated.
+   */
+  private fun verifyAuthState() {
```
oskar-codes, line 33: Consider renaming this to something along the lines of `addAuthStateVerifier`, to specify that this actually creates a new listener, and is not a one-time check.

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
```diff
@@ -53,14 +54,17 @@ import com.google.firebase.Firebase
 import com.google.firebase.auth.auth
 
 @Composable
-fun WelcomeScreen(navigationAction: NavigationAction) {
+fun WelcomeScreen(
+    navigationActions: NavigationAction,
```
oskar-codes, line 58: Consider removing both these parameters as they are not used by the function.

app/src/main/java/com/android/unio/ui/navigation/NavigationAction.kt
```diff
@@ -18,6 +17,9 @@ open class NavigationAction(private val navController: NavHostController) {
    * @param screen The screen to navigate to
    */
   open fun navigateTo(screen: String) {
+    if (navController.currentDestination?.route == screen) {
```
oskar-codes, line 20: This can be rewritten as the shorter `getCurrentRoute() == screen`. Also, it might be useful to add the same check for the TLD `.navigateTo` method.

oskar-codes (approved):
Incredible work on this 🎉
This is a much needed fix for an issue that could have caused us some real trouble. I pointed out a few minor changes you could address, but I gladly approve this pull request.

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
Zafouche: Thanks forgot about these

Zafouche:
> Incredible work on this 🎉 This is a much needed fix for an issue that could have caused us some real trouble. I pointed out a few minor changes you could address, but I gladly approve this pull request.

Thank you! Your points have been taken into account and I have made the necessary changes.

PR #120, opened 2024-10-30
Setup a dependency injection system

Romainhir:
### Setup a dependency injection system, like Hilt to pass global view-models, repositories, etc. to nested components.
Actually, all view-models and repositories are passed as argument when we instanciate a class. This imply a lot of duplicate code and reduce readability. The goal of this task is to remove these dependencies given in parameters to use a dependency injection system, like Hilt. Precisely, this mean :

Remove view-models/repositories dependency passed as argument
Setup the "Hilt" library
Use "Hilt" to manage these global dependencies


app/src/androidTest/java/com/android/unio/HiltTestUtility.kt
Zafouche, line 1: Nice one on this. This can save us trouble for repositories (as interfaces) that are hard to mock.

app/src/androidTest/java/com/android/unio/ui/BottomNavigationTest.kt
Zafouche, line 1: Also thanks for these small improvements. This makes our code more consistent.

app/src/main/java/com/android/unio/model/event/EventListViewModel.kt
Zafouche, line 1: It looks like you did the merge with the main branch wrong. This file should be removed, and replaced by `EventViewModel` everywhere. If I haven't missed something, the only files you have to adapt are in the androidTests.

Zafouche (changes requested):
Tremendous job on this, I know this was hard so I thank you for your work! 🚀

There is an issue with your recent merge with main: you have re-added the file `EventListViewModel.kt`, which was recently deleted. Please see the comment on the [file](https://github.com/SwEnt-Group13/Unio/pull/120#discussion_r1832888888).
I will wait that you resolve this since we are currently working together on-site an approve it as soon as you make a commit about this small "typo"!
Other than that, there really isn't much else to say.

Your Hilt addition goes a long way and makes the usage of view models in our project much cleaner and more efficient!

Zafouche (approved):
Thank you for correcting the merge conflict hiccup!
Also good job on fixing the CI issue. I don't know what the problem was and what the fix was, but that looked like a painful process so thank you 😅

app/src/androidTest/java/com/android/unio/ui/ScreenDisplayingTest.kt
```diff
@@ -64,23 +71,20 @@ class ScreenDisplayingTest {
...
-        {
-          val onSuccess = args[1] as (List<Event>) -> Unit
-          onSuccess(emptyList())
-        }
+    associationViewModel =
+        spyk(
+            AssociationViewModel(
+                mock(), mockk<EventRepositoryFirestore>(), imageRepositoryFirestore))
+    eventListViewModel = spyk(EventListViewModel(mock(), mock()))
+    userViewModel = spyk(UserViewModel(mock()))
```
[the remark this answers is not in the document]
Zafouche: Right, but after having tested spying I haven't found any solution to set the `user` State flow of the user view model to a mock user. And it is indispensable for the user view model to have a non-null user such that the Map screen can work.

PR #122, opened 2024-10-31
Create the Settings screen

oskar-codes:
This pull request includes a new settings screen, accessible from the User Profile's bottom sheet. The settings for now only include the option to change the application's theme and wether or not the app sends notifications. These are only for demonstration purposes and are subject to change, as we add more settings later on.
As such, the main purpose of this pull request is to configure the [ComposePreferences](https://github.com/zhanghai/ComposePreference) library, and use it for two settings mentionned above.
**Important**: Note that these settings are not synced with Firestore, and are instead stored locally in the [SharedPreferences](https://developer.android.com/training/data-storage/shared-preferences). This is intentional, and the settings that are expected to be synced with Firestore across devices will later be available in the Edit Profile screen, not the Settings screen.

Additionally, this pull request has some minor additions that I noticed were needed during the development. These are:
- Include missing fields to the serialization process.
- Add extended test for serialization, to prevent forgetting fields in the serialization function of Users, Associations and Events.
- Refactor the name `followingAssociations` to `followedAssociations` where it was not already refactored.
- Optimize imports project-wide.


oskar-codes:
I'll mark this as ready for review, but to increase the coverage on `MainActivity` we'll need to write end-to-end tests soon.

Zafouche (approved):
Great PR 🚀. There's really not much to say. The code is concise and I haven't spotted any absurdity.
Also thanks for running `Optimize imports`, from the looks of the number of files that were affected, it looks like it's easily forgotten 🙃
Unfortunately I cannot quickly test it out on my emulator, it looks like Firebase is very slow tonight and has some issues (from what I can tell from reading the logs)

PR #93, opened 2024-10-21
Feature/event list v3

Aurelien9Code:
This PR introduces updates to the Home page (also known as the events list). It includes final UI changes to the Event card (for now – I may revisit elements like the "favorite" button later). There's also been extensive refactoring, with class names changed and functions moved for greater consistency.

Special care was taken to ensure that everything is properly mocked so that the Preview functionality works smoothly. The "favorite" button is now functional and integrated with Firebase. **Important**: To make it work, ensure there is a `savedEvent` list in your Firebase database. This list may not exist for users created before the introduction of the event-saving feature.

Additionally, I've split what was once a shared view model for both the EventList and EventCard into two separate view models, one for each.

I’ve also added two specific images to handle image loading issues: a "No image found" placeholder for cases when Firebase fails to fetch an image, and a "Preview image" for mockups in Preview mode.

As for testing, you’ll notice there’s a lack of tests at the moment. I plan to create a level of abstraction for all Firebase-related tests using Mockito, making it easier to mock Firebase actions. Since edge cases are similar across class implementations, the idea is to consolidate these tests into a single file. Adding new tests would simply involve calling this class with specific parameters. This will be my next PR. Thanks for reviewing !

PR #124, opened 2024-10-31
Feature/saved events

Aurelien9Code:
This PR implements saving event page, using saving event management of  #93 !

It basically filters the events from the event list to keep only the saved one. Also, the event are separated between "today"'s event and "upcoming event" category. Also, I made sure that unsaving an event from the Saved page does not immediatly remove it from the page, as it could have been a missclic. Thanks !


Aurelien9Code:
As I merged this PR into my previous event-list v3 PR before merging to main, this one can be closed.

PR #129, opened 2024-11-01
chore:locked screen orientation

armouldr:
Locked screen orientation as it isn't useful for our app. It is just a one line fix in the MainActivity


Aurelien9Code (approved):
Great ! This is a good idea conceptually. However I recommend not to forget the bugs that's it causing when rotating the phone (I think Zafar fixed it but now there is a new bug).

Aurelien9Code:
Also please link this PR to the issue in the scrum board ; )

PR #136, opened 2024-11-07
Add additional settings

oskar-codes:
This pull request adds two additional settings that we need for the app:
- Changing the app language (currently only supports english and french)
- Requesting app location permissions (as requested by @Zafouche to implement the map)

This also comes with a refactor of the settings, in order to make it easier to include additional settings.

Note that this pull request makes use of the [ComposePreference](https://github.com/zhanghai/ComposePreference) library.


app/src/main/java/com/android/unio/model/preferences/AppPreferences.kt
Zafouche, line 1: Good call on renaming this file!

Zafouche, line 1: Good call on renaming this file!

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -52,35 +62,76 @@ fun SettingsScreen(navigationAction: NavigationAction) {
...
+        locationPermissions.launchMultiplePermissionRequest()
+      }
+
+  /** Language * */
+  val language = preferences.get<String>(AppPreferences.LANGUAGE) ?: AppPreferences.Language.default
+  val locale = Locale(language)
+  Locale.setDefault(locale)
+
+  val configuration = context.resources.configuration
+  configuration.setLocale(Locale(language))
```
Zafouche, line 90: Can re-use `val locale = Locale(language)` instead of instantiating the same object twice

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -52,35 +62,76 @@ fun SettingsScreen(navigationAction: NavigationAction) {
...
-                      Theme.DARK -> Icons.Default.Nightlight
-                      Theme.LIGHT -> Icons.Default.WbSunny
+                      AppPreferences.Theme.DARK -> Icons.Default.Nightlight
+                      AppPreferences.Theme.LIGHT -> Icons.Default.WbSunny
                       else -> Icons.Default.Smartphone
                     },
                 contentDescription = context.getString(R.string.settings_theme_content_description))
           })
+      listPreference(
+          modifier = Modifier.testTag(AppPreferences.LANGUAGE),
```
Zafouche, line 119: Should we really use the class's object as a test tag? Perhaps separating test tags from object strings might be a good convention but as it avoids creating more string objects I don't think it's that big of a deal.

Zafouche (approved):
Great work! Thank you for the location request, this will help me center the map on the user's position if they accept it.

Nothing important to point out, LGTM.

app/src/main/java/com/android/unio/model/preferences/AppPreferences.kt
```diff
@@ -0,0 +1,44 @@
+package com.android.unio.model.preferences
+
+object AppPreferences {
+  const val THEME = "theme"
+  const val NOTIFICATIONS = "notifications"
+  const val LOCATION_PERMISSION = "location_permission"
+  const val LANGUAGE = "language"
+
+  object Theme {
```
Redd87, line 9: Could it be useful to also add the options to change the contrast of the application if we also allow the user to change the theme here?

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -52,35 +62,76 @@ fun SettingsScreen(navigationAction: NavigationAction) {
...
+  val language = preferences.get<String>(AppPreferences.LANGUAGE) ?: AppPreferences.Language.default
+  val locale = Locale(language)
+  Locale.setDefault(locale)
+
+  val configuration = context.resources.configuration
+  configuration.setLocale(Locale(language))
+  configuration.setLayoutDirection(locale)
+
+  context.createConfigurationContext(configuration)
+  context.resources.updateConfiguration(configuration, context.resources.displayMetrics)
```
Redd87, line 94: Why use the deprecated method updateConfiguration?

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -52,35 +62,76 @@ fun SettingsScreen(navigationAction: NavigationAction) {
...
 
   ProvidePreferenceLocals(flow = LocalPreferenceFlow.current) {
     LazyColumn(
         modifier = Modifier.testTag("SettingsContainer"),
     ) {
       listPreference(
```
Redd87, line 100: Consider adding the link to the documentation of the library you use in the comment since it is a somewhat custom solution

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -90,6 +141,29 @@ fun SettingsContainer() {
...
+          summary = {
+            if (locationPermissions.allPermissionsGranted) {
+              Text(context.getString(R.string.settings_location_summary_granted))
+            } else {
+              Text(context.getString(R.string.settings_location_summary_denied))
+            }
+          },
+          icon = {
+            Icon(
+                imageVector = Icons.Default.LocationOn, contentDescription = "Location Permissions")
```
Redd87, line 157: You might want to move the string of contentDescription to the file strings.xml, I think under the accessbility and content description

Redd87 (approved):
A strong implementation of the settings! Everything I found was minor, the actual functionality of the code looks good to me

app/src/main/java/com/android/unio/ui/settings/Settings.kt
oskar-codes: This is needed to support older versions of Android.

app/src/main/java/com/android/unio/ui/settings/Settings.kt
oskar-codes: Good catch!

app/src/main/java/com/android/unio/model/preferences/AppPreferences.kt
oskar-codes: Indeed, if needed, other themes can be easily added with this preference refactor.

PR #139, opened 2024-11-07
Feat/map integrate Firestore database for saved events

Zafouche:
## Description:

This PR integrates Firestore database usage for retrieving and displaying saved events within the MapScreen. This means replacing the previous arbitrary mock saved event with the list of fetched saved events from Firestore. The saved event list is fetched through the user view model. This means that if the user saves an event, then the map composable will refresh and update its list of saved events.

## Motivation and context:

Previously, the MapScreen was using arbitrary mock data to represent saved events, which was only temporary while waiting for the "saved events" logic to be available to me, which was being implemented by another teammate. This update PR allows the app to fetch saved events associated with the user on Firestore. This data is dynamically updated, which means the markers have also been adapted to be updated dynamically.

### MapScreen Integration:

- Replaced the mock list with the actual saved events fetched from the Firestore database.
- Integrated the view model such that the markers' list is updated and the markers refreshed when user data is updated (e.g. user just saved an event).
- Special care was taken into merging updates from main such that failing tests pass asap.

## How has this been tested?

The integration has been tested using the existing UI tests and unit tests, additional tests don't seem necessary yet.

## Note:

I will work on implementing the display of events near the user based on location permissions, e.g. center the map on the user if the user allows it, when @oskar-codes publishes his PR which is about location permission requests and it makes it onto the main branch. 

### Edit:
After some consideration, this last addition will be done as soon as the next sprint starts as I need to document myself on how to get the GPS location from the phone's system.


Zafouche:
Working on the CI failure (works locally).

Zafouche:
I find it interesting, that sometimes, updates from main onto the branch automatically overwrites and "cancels" previous additions I had done. I guess GitHub has its flaws after all.

oskar-codes (approved):
This pull request is proof that you don't need a lot of code to make impactful contributions 💪 Great work on this as always, @Zafouche.

Zafouche:
> This pull request is proof that you don't need a lot of code to make impactful contributions 💪 Great work on this as always, @Zafouche.

Danke schön brother!

PR #142, opened 2024-11-07
Chore: refactor tests and cleanup view models

Zafouche:
## Description:

This PR addresses code quality improvements and refactoring in response to feedback received from the first sprints. The updates contribute to a more consistent code.

## Motivation and context:

During the first two sprints, certain coding issues were noted by the coaches. This PR aims to address the following remarks:

- Commented-out tests in the code. If these tests aren’t passing due to specific issues, consider adding a note to explain why they’re currently disabled. If they’re not useful, however, it’s best to remove them entirely to keep the codebase clear and reduce clutter.
- A few instances of hardcoded text were found; it would be best to extract these and place them in the string resources file for maintainability.
- In AssociationViewModel, asynchronous launch with a coroutine was used, but the code is structured around callbacks rather than coroutines. Sticking with one approach is recommended; in this case, removing the coroutine launch would be more efficient.
- In the view models, when using a MutableStateFlow to back a StateFlow, remember to call asStateFlow() on the MutableStateFlow. This will create a readOnlyStateFlow, making it clear that the value is immutable and preventing modifications from outside the viewModel.
- There’s an opportunity to improve test coverage, particularly around UI and viewModel code, which are generally straightforward to test and would enhance reliability.

### Code Cleanup:

- Removed unnecessary commented-out code. **(TBD)**
- Relocated hardcoded text to strings.xml. **(TBD / To check if not already done)**

### AssociationViewModel:

- Removed Factory companion object
- Removed coroutines

### StateFlow Refactoring:

- Applied asStateFlow() on MutableStateFlow properties in the view models to enforce immutability **(TBD)**

### Test Coverage:

- Expanded test cases for UI components and view models to enhance code reliability **(TBD / To check if not already done)**

PR #157, opened 2024-11-11
Chore/refactor tests and viewmodel cleanup part 2

Zafouche:
This is the continuation of the work done in #142.

## Description:

This PR addresses code quality improvements and refactoring in response to feedback received from the first sprints. The updates contribute to a more consistent code and will allow us to have a strong base for the project such that we can carry out good practices easily onwards.

## Motivation and context:

During the first two sprints, certain coding issues were noted by the coaches. This PR aims to address the following remarks:

- Commented-out tests in the code. If these tests aren’t passing due to specific issues, consider adding a note to explain why they’re currently disabled. If they’re not useful, however, it’s best to remove them entirely to keep the codebase clear and reduce clutter.
- A few instances of hardcoded text were found; it would be best to extract these and place them in the string resources file for maintainability.
- In AssociationViewModel, asynchronous launch with a coroutine was used, but the code is structured around callbacks rather than coroutines. Sticking with one approach is recommended; in this case, removing the coroutine launch would be more efficient.
- In the view models, when using a MutableStateFlow to back a StateFlow, remember to call asStateFlow() on the MutableStateFlow. This will create a readOnlyStateFlow, making it clear that the value is immutable and preventing modifications from outside the viewModel.
- There’s an opportunity to improve test coverage, particularly around UI and viewModel code, which are generally straightforward to test and would enhance reliability.

The goal of this PR is to address issues like the above generally in the project.

### Code Cleanup:

- Removed unnecessary commented-out code and tests. **(Done in pt1)**
- Relocated hardcoded text to strings.xml. **(Done)**
- Relocated hardcoded test tags to the model/strings/test_tags folder. **(Done)**

### View models:

- Removed Factory companion objects. **(Done in pt1)**
- Removed coroutines. **(Done in pt1)**

### StateFlow Refactoring:

- Applied asStateFlow() on MutableStateFlow properties in the view models to enforce immutability **(Done in pt1)**

### Test Coverage:
Expanded test cases for view models and repositories **(Done)**
- Added tests for the `init` functions of `AssociationRepository` and UserRepository.
- Added tests for the uncovered functions of `AssociationViewModel`, e.g. `getEventsForAssociation` and `saveAssociation` (which was barely covered)


Zafouche:
For the view model testing: 
The view models look pretty well tested, except for one: UserViewModel. If I remember correctly, @Aurelien9Code , you had added additional functions to the base of the view model @oskar-codes had implemented. However Oskar and me have raised a concern: the defined functions could simply be not defined in the view model, as they are only used in EventCard, and consist of some simple if else logic. So I suggest, either you add tests for them, or keep the logic in the EventCard file. Let us know what you think about this concern.

Zafouche:
Added tests for the uncovered functions of `AssociationViewModel`, e.g. `getEventsForAssociation` and `saveAssociation` (barely covered). Opening for review.

oskar-codes (approved):
You're doing some real heavy lifting with this pull request! Nothing to say, thanks for the incredible work @Zafouche 🥇

Zafouche:
> You're doing some real heavy lifting with this pull request! Nothing to say, thanks for the incredible work @Zafouche 🥇

Alt Text

PR #158, opened 2024-11-11
Feature/event creation frontend partial implementation

Redd87:
This PR partially implements the event creation front end by an association. 

## Description
It adds a dummy Create an event button which is only shown if the variable isAdmin = true, which is always the case in this PR. It will later be modified to allow a user view and an admin view of an association, this simply sets the groundwork. 
The event creation page is implemented as shown in the Figma; it simply consists of a text inputs, an image picker and a custom date picker which allows to set a beginning and end of an event. As this pull request was getting quite large already, I didn't implement the location picker and the coorganizing associations picker. 
To allow for navigation I modified the signature of the EventCreationScreen and the places that call it as required.

### Tests
The elements of the event creation page are tested to be displayed in the EventCreationTest.kt file, but a full functionality test that checks that inputs are valid will be done alongside the location and association picker.

### TBD
Add remaining frontend functionality, input validation checks and connect to the create button to the database


app/src/main/java/com/android/unio/model/strings/test_tags/EventTestTags.kt
Zafouche, line 1: Thank you for adding the event test tags.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -448,6 +450,18 @@ private fun AssociationEvents(
...
+  if (isAdmin) {
+    Button(
+        onClick = { navigationAction.navigateTo(Screen.EVENT_CREATION) },
+        contentPadding = ButtonDefaults.ButtonWithIconContentPadding) {
+          Icon(
+              Icons.Filled.Add,
+              contentDescription = "Add event",
+              modifier = Modifier.size(ButtonDefaults.IconSize))
+          Spacer(Modifier.size(ButtonDefaults.IconSpacing))
+          Text("Add event")
```
Zafouche, line 462: Hardcoded strings!!!!! 😡😡😡😡😡😡

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+  var shortDescription by remember { mutableStateOf("") }
+  var longDescription by remember { mutableStateOf("") }
+
+  Scaffold(
+      modifier = Modifier.testTag(EventCreationTestTags.SCREEN),
+      content = { padding ->
+        Column(
+            modifier =
+                Modifier.padding(padding)
+                    .padding(horizontal = 20.dp, vertical = 20.dp)
```
Zafouche, line 85: This can be simplified to `.padding(20.dp)`

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+                    .fillMaxWidth()
+                    .verticalScroll(rememberScrollState()),
```
Zafouche, line 87: Maybe you should extract `rememberScrollState`, I am not sure but I think it creates a remember each refresh, since it isn't stored in a variable.

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+                awaitFirstDown(pass = PointerEventPass.Initial)
+                val upEvent = waitForUpOrCancellation(pass = PointerEventPass.Initial)
+                if (upEvent != null) {
+                  isTimePickerVisible = true
+                }
+              }
+            },
+        value = selectedTime?.let { convertMillisToTime(it) } ?: "",
+        readOnly = true,
+        onValueChange = {},
```
Zafouche, line 278: Can this line be removed?

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+        trailingIcon = { Icon(Icons.Default.AccessTime, contentDescription = "Select date") },
```
Zafouche, line 279: Hardcoded String!!!!!!!!! 😡 😡 😡 😡 😡 😡 😡 😡 😡 😡 😡 😡 😡

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun TimePickerModal(onTimeSelected: (Long?) -> Unit, onDismiss: () -> Unit) {
+  val timePickerState = rememberTimePickerState(is24Hour = true)
+
+  TimePickerDialog(
+      onDismiss = onDismiss,
+      onConfirm = {
+        onTimeSelected(
+            timePickerState.hour * 60 * 60 * 1000L - 3600000 + timePickerState.minute * 60 * 1000L)
```
Zafouche, line 338: Consider using `TimeUnit` to make these conversions more readable, e.g. to go from an hour value to a ms value, use `TimeUnit.HOURS.toMillis()`

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -1,20 +1,379 @@
...
+}
+
+fun convertMillisToDate(millis: Long): String {
+  val formatter = SimpleDateFormat("dd/MM/yy", Locale.getDefault())
+  return formatter.format(Date(millis))
+}
+
+fun convertMillisToTime(millis: Long): String {
+  val formatter = SimpleDateFormat("HH:mm", Locale.getDefault())
+  return formatter.format(Date(millis))
```
Zafouche, line 378: If I remember correctly @Aurelien9Code had written some similar code for time -> date conversion somewhere. We should consider extracting these functions in an utility file.

Zafouche (approved):
Good work @Redd87! Not much to say.
I can understand why you cannot test this any further as this screen does not implement any functionality yet.
Other than that I have pointed out some details like forgotten hardcoded strings, so please take care of those before merging.

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
Redd87: No this is sadly boilerplate that has to be kept as it doesn't have a default value

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
Zafouche: Don't think we do. It's alright we can do it in a different PR later since it should only concern one other file

PR #155, opened 2024-11-11
Add [End to end Test] New end to end tests for authentication with Firebase Emulator Suite

AlouchLaBouche:
This PR adds the a new end to end test for the login and account creation process.
This Test uses the Firebase Emulator suite, an emulated Firebase server but that is run locally on your device. Please read the [wiki](https://github.com/SwEnt-Group13/Unio/wiki/Firebase-Emulator-Tutorial) to understand the motivation and how to use it yourselves.

This PR includes: 
- A new `EndToEndTest.kt` file in the authentication package. This is the main test that looks at the behavior of the login and account creation process.
- Additions to `ci.yml`to include the startup of the emulators, and the execution of our new test.
- New configuration files that are created by default by when initializing the Firebase emulator (the `rules` for `Firebase` and `storage` are the same that are on the main unio production Database). Other files are all initialized by default.

More details about this PR are provided with the linked [issue](https://github.com/SwEnt-Group13/Unio/issues/134), the [wiki](https://github.com/SwEnt-Group13/Unio/wiki/Firebase-Emulator-Tutorial) and in the provided comments to help with the code review, because 7000 lines does seem a bit intimidating :sweat_smile: .

PR #161, opened 2024-11-11
Create Firestore security rules

oskar-codes:
This pull request aims to create baseline security rules for the Firestore database. Currently, the database gives read/write access to any authenticated user.
The new rules provided in this pull request enforce authenticated requests, and validate the uploaded data. Note that the rules I wrote prefer data integrity over confidentiality.

Feel free to review the rules and give opinions, as they are subject to change and need to be discussed with the entire team.

### DIFF
--- a/firebase/tests/firestore-mock-data.js
+++ b/firebase/tests/firestore-mock-data.js
@@ -1,0 +1,111 @@
+import { setDoc, doc } from "firebase/firestore";
+
+export const alice = {
+  uid: "alice",
+  email: "alice@wonderland.com",
+  firstName: "Alice",
+  lastName: "Wonderland",
+  biography: "I'm a curious",
+  savedEvents: [],
+  followedAssociations: ["other-association"],
+  joinedAssociations: ["alice-association"],
+  interests: [],
+  socials: [],
+  profilePicture: "",
+};
+
+export const aliceAssociation = {
+  uid: "alice-association",
+  url: "alice-association",
+  name: "AA",
+  fullName: "Alice Association",
+  category: "UNKNOWN",
+  description: "Description",
+  followersCount: 0,
+  members: ["alice"],
+  image: "",
+  events: [],
+};
+
+export const aliceEvent = {
+  uid: "event",
+  title: "Event",
+  organisers: ["alice-association"],
+  taggedAssociations: [],
+  image: "",
+  description: "Description",
+  catchyDescription: "Catchy description",
+  price: 0.0,
+  date: new Date(),
+  location: {
+    name: "Location",
+    address: "Address",
+    latitude: 0.0,
+    longitude: 0.0,
+  },
+  types: ["OTHER"],
+};
+
+export const otherUser = {
+  uid: "other",
+  email: "other@unknown.com",
+  firstName: "Other",
+  lastName: "Person",
+  biography: "I'm a curious",
+  savedEvents: [],
+  followedAssociations: ["alice-association"],
+  joinedAssociations: ["other-association"],
+  interests: [],
+  socials: [],
+  profilePicture: "",
+}
+
+export const otherEvent = {
+  uid: "other-event",
+  title: "Other Event",
+  organisers: ["other-association"],
+  taggedAssociations: [],
+  image: "",
+  description: "Description",
+  catchyDescription: "Catchy description",
+  price: 0.0,
+  date: new Date(),
+  location: {
+    name: "Location",
+    address: "Address",
+    latitude: 0.0,
+    longitude: 0.0,
+  },
+  types: ["OTHER"],
+};
+
+export const otherAssociation = {
+  uid: "other-association",
+  url: "other-association",
+  name: "OA",
+  fullName: "Other Association",
+  category: "UNKNOWN",
+  description: "Description",
+  followersCount: 0,
+  members: ["other"],
+  image: "",
+  events: [],
+};
+
+export async function setupFirestore(testEnv) {
+  testEnv.clearFirestore();
+
+  console.log("Setting up Firestore data...");
+  await testEnv.withSecurityRulesDisabled(async env => {
+    const db = env.firestore();
+
+    await setDoc(doc(db, `/users/${alice.uid}`), alice);
+    await setDoc(doc(db, `/events/${aliceEvent.uid}`), aliceEvent);
+    await setDoc(doc(db, `/associations/${aliceAssociation.uid}`), aliceAssociation);
+
+    await setDoc(doc(db, `/users/${otherUser.uid}`), otherUser);
+    await setDoc(doc(db, `/associations/${otherAssociation.uid}`), otherAssociation);
+    await setDoc(doc(db, `/events/${otherEvent.uid}`), otherEvent);
+  });
+  console.log("Done.");
+}
--- a/firestore.rules
+++ b/firestore.rules
@@ -1,0 +1,115 @@
+rules_version = '2';
+
+// Read rules can be divided into get and list rules
+// Write rules can be divided into create, update, and delete rules
+
+service cloud.firestore {
+  match /databases/{database}/documents {
+
+    function isSignedIn() {
+      return request.auth != null;
+    }
+    function isVerified() {
+      return isSignedIn() && request.auth.token.email_verified;
+    }
+
+    match /users/{uid} {
+      function isOwner() {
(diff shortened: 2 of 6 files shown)


storage.rules
```diff
@@ -0,0 +1,12 @@
...
+// Craft rules based on data in your Firestore database
+// allow write: if firestore.get(
+//    /databases/(default)/documents/users/$(request.auth.uid)).data.isAdmin;
+service firebase.storage {
+  match /b/{bucket}/o {
+    match /{allPaths=**} {
+      allow read, write: if false;
+    }
+  }
+}
```
Redd87, line 12: What is this file?

Redd87:
I briefly went through the rules and I agree with them,  though I'll still need to do an actual review. I agree with the association creation request architecture. There won't be that many, it's better to moderate them by going through them manually rather trying to create an in-app moderation system that restrains users from create a false balelec association

storage.rules
oskar-codes: Those are the default security rules of the Firebase Storage service, handling for example our images etc. This pull request does not consider those, I only worked on the Firestore rules.

package-lock.json
AlouchLaBouche, line 1: Watch out, this file will be conflicting with my PR #155. This might cause some nasty conflicts with the slightest conflicts with the file (and they are different as mine is 7000 lines long :sweat: ). So let's agree for one of us to push it.

firebase.json
AlouchLaBouche, line 1: Why are there 'Host' values here, are they generated automatically. And shouldn't they be "10.0.2.2" ?

package-lock.json
oskar-codes: Dont't worry, package-lock.json files will not cause any trouble as they are auto generated by NPM. We can just delete both files when merging and the run `npm install`.

firebase.json
oskar-codes: The `"host"` properties in this file specify which host the emulators should start on. They should be set to `127.0.0.1` for all tests, including Android tests as that's where the Firebase Emulators will run on the computer. The device emulator can then access them using `10.0.2.2` (which redirects to `127.0.0.1` from the emulated device), but the Firebase emulators will always be running on `127.0.0.1`.

I added this property because the CI failed to correctly run the emulators.

firebase/tests/firestore-mock-data.js
AlouchLaBouche, line 1: Nice extraction of constant values :+1: .

firebase.json
AlouchLaBouche, line 1: Are the "host" fields necessary here and is the IP local host correct, shouldn't it be "10.0.2.2"

.github/workflows/ci.yml
AlouchLaBouche, line 1: Nice modification of the CI workflow 👍 . Separating the rules check into another "task" is a good idea.

firebase/tests/firestore-rules.test.js
AlouchLaBouche, line 1: The tests look flawless to me :+1:.

firestore.rules
```diff
@@ -0,0 +1,129 @@
...
+        // Prevent the user from changing the joinedAssociations field
+        // This will be handled by a Firebase Function
+        return request.resource.data.joinedAssociations == resource.data.joinedAssociations;
+      }
+      function validate() {
+        let fields = ['uid', 'email', 'firstName', 'lastName', 'biography', 'savedEvents', 'followedAssociations', 'joinedAssociations', 'interests', 'socials', 'profilePicture'];
+        let hasCorrectNumberOfFields = request.resource.data.size() == fields.size();
+        let hasCorrectFields = request.resource.data.keys().hasAll(fields);
+
+        return hasCorrectNumberOfFields && hasCorrectFields &&
```
AlouchLaBouche, line 30: This boolean is quite difficult to understand. What is it doing ? maybe add a comment or extract it to a function that has a comprehensible name?

firestore.rules
```diff
@@ -0,0 +1,129 @@
...
+        let onlyFollowerCountChanged = affectedKeys.hasOnly(['followersCount']);
+
+        return onlyFollowerCountChanged && (newCount == oldCount || newCount == oldCount + 1 || newCount == oldCount - 1);
+      }
+      function validate() {
+        let fields = ['uid', 'url', 'name', 'fullName', 'category', 'description', 'followersCount', 'members', 'image', 'events'];
+        let hasCorrectNumberOfFields = request.resource.data.size() == fields.size();
+        let hasCorrectFields = request.resource.data.keys().hasAll(fields);
+
+        return hasCorrectNumberOfFields && hasCorrectFields &&
```
AlouchLaBouche, line 68: Same as above, I believe this boolean deserves it's own function.

firebase.json
oskar-codes: See [this](https://github.com/SwEnt-Group13/Unio/pull/161/files/9f60f091befc1213a43d1d5a3198bd486b568605#r1841909245) comment.

firestore.rules
oskar-codes: This entire expression does data validation of a user object. That means that it checks that the uploaded object contains exactly the specific fields defined above, and each field is of the correct type.
These can be improved at a later point in time with even more data validation, for example checking that `profilePicture` is an email coming from our storage bucket, but for now this seems okay to me.

Also, note that Firestore restricts the amount of functions/variables declared in a ruleset in order to limit their processing time. Thus, I believe we should not extract too many expressions in their own functions/variables in order to allow further scaling. See the [Firestore security rules limits](https://firebase.google.com/docs/firestore/security/rules-structure#security_rule_limits).

firebase/tests/firestore-rules.test.js
Aurelien9Code, line 1: Wow very strong implementation ! 🚀

However, you should test edge cases such as empty fields, null values, and large data sets to ensure the rules handle all scenarios. Also, you may need to ensure the mock data covers all fields expected by the Firestore rules, especially if new fields are added. Also I saw that there is not a lot of tests where multiple conditions or rules interact, e.g., updating both followersCount and members simultaneously. In the end, you functions seem a bit too big.

Aurelien9Code (approved):
The implementation is solid !!! You made comprehensive tests covering most scenarios. To be even more robust, consider addressing edge cases, expanding mock data coverage and testing more complex permission interactions. You could also moduralize the tests for better scalability. But nothing that needs to be done right now ;)

firestore.rules
Aurelien9Code: Oh sorry I didn't see that. Of course then doing helper functions might not be needed if there's a limit

Aurelien9Code (approved):

firestore.rules
```diff
@@ -0,0 +1,135 @@
...
+    }
+    function isVerified() {
+      return isSignedIn() && request.auth.token.email_verified;
+    }
+
+    match /users/{uid} {
+      function isOwner() {
+        return isVerified() && request.auth.uid == uid;
+      }
+      function validateJoinedAssociations() {
```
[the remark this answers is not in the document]
oskar-codes: Indeed, I added basic data validation in the `validate` function that only checks data types, but this is intended to be built upon as this pull request only provides baseline security rules.

firestore.rules
```diff
@@ -0,0 +1,135 @@
...
+      function validate() {
+        // CHeck that the user document has the correct fields
+        let fields = ['uid', 'email', 'firstName', 'lastName', 'biography', 'savedEvents', 'followedAssociations', 'joinedAssociations', 'interests', 'socials', 'profilePicture'];
+        let hasCorrectNumberOfFields = request.resource.data.size() == fields.size();
+        let hasCorrectFields = request.resource.data.keys().hasAll(fields);
+
+        return hasCorrectNumberOfFields && hasCorrectFields &&
+          request.resource.data.uid == request.auth.uid &&
+          request.resource.data.email == request.auth.token.email &&
+          request.resource.data.firstName is string &&
```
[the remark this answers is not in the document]
oskar-codes: The values are already asserted to be their respective type, which thus does not allow null and undefined values.

firestore.rules
```diff
@@ -0,0 +1,135 @@
...
+          request.resource.data.lastName is string &&
+          request.resource.data.biography is string &&
```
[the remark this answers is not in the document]
oskar-codes: See [this](https://github.com/SwEnt-Group13/Unio/pull/161#discussion_r1842135576) comment.

PR #164, opened 2024-11-13
Feat/map: user gps location

Zafouche:
## Description

This PR introduces GPS location usage for the `MapScreen`, enabling the map to center on the user’s location when opened (It first checks for the user's authorized permissions for location, and asks for them one time if they are not activated). I have added a `MapViewModel` to handle the GPS data, which is used in `MapScreen` to implement the GPS.
So when first opening the Map screen, the user is prompted with location permissions requests (only 1 time). 

If the user has allowed precise location, then the map is starts on the user's precise location, with a blue point resembling the one seen on Google Maps.
If the user has only allowed approximate location, the the map is starts on the user's approximate location, with a blue circle of significant size, to display the approximation.
If the user refuses any permissions for the GPS, then the map starts at EPFL's location.

## Motivation and context

This update allows us to use the GPS in our app. The map implements the live GPS. In the future, this might allow us to add a filter to show events by proximity to their current location.

### MapViewModel
- Created a view model to manage the location of the user.
- The view model uses the FusedLocationProviderClient made available by android.

### MapScreen
- Added permission prompts at the first time opening of the map screen.
- Implemented logic that checks which permissions have been allowed by the user, which ultimately decides the map's behavior (described in the `Description` section above)
- Integrated the `MapViewModel` to manage the GPS data of the user.
- Implemented a centering FAB in MapScreen to focus on the user’s location, if permissions are enabled and a location is fetched.
- Added functionality to display the user’s pin on the map similar to Google Maps, if the permissions are enabled.

## How has this been tested?

- MapViewModel: The behavior of the view model has been fully tested, meaning if the permissions are enabled or not, and if the fusedLocationProviderClient fails to fetch a location and throws an exception.
- MapScreen: Mocked the map view model and the fused location provider and tested basic map display.


Zafouche:
Testing the permission requests is hard  (@oskar-codes hasn't been able to do it either), and as I have already spent too much time on this, and it is not a priority, this will potentially be done in a later PR. Opening this PR for review shortly.

app/src/androidTest/java/com/android/unio/ui/ScreenDisplayingTest.kt
```diff
@@ -120,6 +142,29 @@ class ScreenDisplayingTest {
     eventViewModel.loadEvents()
     eventViewModel.selectEvent(events.first().uid)
 
+    // Mocking the location permission checks
+    instrumentation = InstrumentationRegistry.getInstrumentation()
+    instrumentation.uiAutomation.executeShellCommand(
+        "pm grant ${InstrumentationRegistry.getInstrumentation().targetContext.packageName} android.permission.ACCESS_FINE_LOCATION")
+    instrumentation.uiAutomation.executeShellCommand(
+        "pm grant ${InstrumentationRegistry.getInstrumentation().targetContext.packageName} android.permission.ACCESS_COARSE_LOCATION")
```
armouldr, line 150: How does this work and why didn't you do the same in MapScreenTest ? I'm not sure to understand how these permission tests work.

app/src/androidTest/java/com/android/unio/ui/map/MapScreenTest.kt
armouldr, line 1: Good tests that are pretty thorough ! Especially the `centerOnUserFabCentersMap` one

app/src/main/java/com/android/unio/model/map/MapViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+
+  /** Fetches the user's location and updates the [_userLocation] state flow. */
+  fun fetchUserLocation(context: Context) {
+    if (hasLocationPermissions(context)) {
+      try {
+        fusedLocationClient.lastLocation.addOnSuccessListener { location ->
+          location?.let { _userLocation.value = LatLng(it.latitude, it.longitude) }
+        }
+      } catch (e: SecurityException) {
+        Log.e("MapViewModel", "Permission for location access was revoked: ${e.localizedMessage}")
```
armouldr, line 33: Good error handling

app/src/main/java/com/android/unio/ui/map/Map.kt
armouldr, line 1: This can be implemented in a future PR, but as of now the location is only fetched one time when the screen is loaded, and is thus not updated in real time. Tell me if I am wrong but that is what I understand from the code.
Also, maybe you should create constants for the different zoom levels `10f` and `15f`

app/src/main/java/com/android/unio/ui/map/Map.kt
```diff
@@ -16,35 +23,115 @@ import androidx.compose.runtime.Composable
...
+        CameraPosition.fromLatLngZoom(userLocation ?: EPFL_COORDINATES, 10f)
+  }
+
+  // Center map on the user's location initially if available
+  LaunchedEffect(userLocation) {
+    if (userLocation != null && !initialCentered) {
+      cameraPositionState.position = CameraPosition.fromLatLngZoom(userLocation!!, 15f)
+      initialCentered = true
+    }
+  }
```
armouldr, line 133: For now this is redundant, as userLocation is only updated when requesting permissions.

app/src/main/java/com/android/unio/ui/theme/Color.kt
```diff
@@ -226,3 +226,7 @@ val eventTypeNetworking = Color(0xFF009688) // Cyan
 val eventTypeSport = Color(0xFF8BC34A) // Light Green
 val eventTypeTrip = Color(0xFFE91E63) // Pink
 val eventTypeOther = Color.Gray // Default Color
+
+// Colors for the approximate user location circle on the map
+val mapUserLocationCircleFiller = Color(0x2200A0FF)
+val mapUserLocationCircleStroke = Color(0xAA00A0FF)
```
armouldr, line 232: It's great that you thought about putting them in Color.kt :)

app/src/test/java/com/android/unio/model/map/MapViewModelTest.kt
```diff
@@ -0,0 +1,106 @@
...
+        {
+          (it.invocation.args[0] as OnFailureListener).onFailure(
+              SecurityException("Security exception"))
+          locationTask
+        }
+
+    mapViewModel.fetchUserLocation(context)
+
+    // Required for the location Task to complete
+    shadowOf(Looper.getMainLooper()).idle()
```
armouldr, line 89: Great find, I didn't know this method existed

armouldr (commented):
Good addition, we finally use one of the phone's sensors :tada: 
I don't have much to say about this as the code is really clean, simply please have a look at the possibly redundant LaunchedEffects before I can approve.

app/src/androidTest/java/com/android/unio/ui/ScreenDisplayingTest.kt
Zafouche: This is exactly the same as in the rule in the MapScreenTest. But you are right, I should remove this and also put the rule there.

app/src/main/java/com/android/unio/ui/map/Map.kt
Zafouche: You're right there probably should be periodical calls to the `MapViewModel`, since the `FusedLocationProviderClient` should already be updated regularly. I'll have to look into asap next sprint.

app/src/main/java/com/android/unio/ui/map/Map.kt
Zafouche: You're right, but it won't be when periodic updates of the view model are introduced next week.

app/src/test/java/com/android/unio/model/map/MapViewModelTest.kt
Zafouche: `Robolectric` is pretty great! Shame it doesn't work for android tests.

armouldr (approved):
LGTM with the changes :+1:

Zafouche:
> LGTM with the changes 👍

Thank you! Ultimately this PR needs improvement, but it brings our first actual usage of a phone sensor (except the camera for pictures).

PR #165, opened 2024-11-14
Feat/Home: Change Home page top bar menu

armouldr:
The top bar menu was previously hardcoded a looked a bit "drafty" so I refactored it. It now uses the TabRow composable. I chose to use a custom indicator lambda so that the indicator moves along with the user's scroll instead of moving after the scroll.
This change took me way too much time but it looks cool ;)

I also cut the Home screen into smaller composables to make it more readable.

### DIFF
--- a/app/src/main/java/com/android/unio/ui/home/Home.kt
+++ b/app/src/main/java/com/android/unio/ui/home/Home.kt
@@ -1,5 +1,6 @@
 package com.android.unio.ui.home
 
+import android.util.Log
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
@@ -197,8 +198,18 @@ fun TopBar(
               Box(
                   modifier =
                       Modifier.fillMaxSize().drawBehind {
-                        val tabWidth = sizeList[0]!!.first
-                        val height = sizeList[0]!!.second
+                        val tabWidth: Float
+                        val height: Float
+                        if (sizeList[0] == null) {
+                          Log.e(
+                              "Home Page", "The size values of tabs are null, should not happen !")
+                          // hardcoded values in case sizeList[0] is null
+                          tabWidth = 576.0F
+                          height = 92.0F
+                        } else {
+                          tabWidth = sizeList[0]!!.first
+                          height = sizeList[0]!!.second
+                        }
                         val startOffset =
                             Offset(
                                 x = progressFromFirstPage * tabWidth + tabWidth / 3,
--- a/app/src/main/java/com/android/unio/ui/home/Home.kt
+++ b/app/src/main/java/com/android/unio/ui/home/Home.kt
@@ -1,6 +1,5 @@
 package com.android.unio.ui.home
 
-import androidx.compose.animation.core.Animatable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
+
+        val progressFromFirstPage by remember {
+          derivedStateOf { pagerState.currentPageOffsetFraction + pagerState.currentPage.dp.value }
+        }
+        val colorScheme = MaterialTheme.colorScheme
+
+        // Tab Menu at the top of the page
+        TabRow(
+            selectedTabIndex = pagerState.currentPage,
+            contentColor = Color(0xFF362C28),
```
AlouchLaBouche, line 194: Hard-coded color ? Won't we have issues with this when changing the app's color theme ?

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
+fun TopBar(
+    navigationAction: NavigationAction,
+    searchState: SearchViewModel.Status,
+    searchResults: List<Event>,
+    userViewModel: UserViewModel,
+    searchViewModel: SearchViewModel,
+    eventViewModel: EventViewModel,
+    paddingValues: PaddingValues
+) {
+  val nbOfTabs = 2
```
AlouchLaBouche, line 168: Maybe extract this constant ? I'm being picky.

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
-            } else {
-              LazyColumn(
-                  contentPadding = PaddingValues(vertical = 8.dp),
-                  modifier = Modifier.fillMaxSize().padding(horizontal = 32.dp)) {
-                    items(searchResults) { event ->
-                      EventCard(navigationAction, event, userViewModel, eventViewModel)
+                  modifier =
+                      Modifier.fillMaxSize().drawBehind {
+                        val tabWidth = sizeList[0]!!.first
+                        val height = sizeList[0]!!.second
```
Romainhir, line 202: Instead of asserting non null, you can add a default value. Because you cannot ensure that this list have non-null element

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
+            inputField = {
+              SearchBarDefaults.InputField(
+                  modifier = Modifier.testTag(HomeTestTags.SEARCH_BAR_INPUT),
+                  query = searchQuery,
+                  onQueryChange = {
+                    // Search when query changes
+                    searchQuery = it
+                    searchViewModel.debouncedSearch(it, SearchViewModel.SearchType.EVENT)
+                  },
+                  onSearch = {},
```
Romainhir, line 254: Is it mandatory to define onSearch ? If not, it can be removed

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
-                  modifier = Modifier.fillMaxSize().padding(horizontal = 32.dp)) {
-                    items(searchResults) { event ->
-                      EventCard(navigationAction, event, userViewModel, eventViewModel)
+                  modifier =
+                      Modifier.fillMaxSize().drawBehind {
+                        val tabWidth = sizeList[0]!!.first
+                        val height = sizeList[0]!!.second
+                        val startOffset =
+                            Offset(
+                                x = progressFromFirstPage * tabWidth + tabWidth / 3,
```
AlouchLaBouche, line 205: The fact that the height is hard-coded might cause some problems with different devices and different screen sizes no ? Somehow make it flexible depending for each device ? I don't know if this is really a problem, maybe the fact it is in a composable function, it places it relatively to the parent component ?

AlouchLaBouche (commented):
In general everything looks good to me and this refactor does make the `Home.kt` file much more readable :rocket: ! Please do consider using Hilt now that @Romainhir has done the hard part of implementing it :+1:

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
+                  modifier = Modifier.testTag(HomeTestTags.SEARCH_BAR_INPUT),
+                  query = searchQuery,
+                  onQueryChange = {
+                    // Search when query changes
+                    searchQuery = it
+                    searchViewModel.debouncedSearch(it, SearchViewModel.SearchType.EVENT)
+                  },
+                  onSearch = {},
+                  expanded = false,
+                  onExpandedChange = {},
```
Romainhir, line 256: Same for this, if not mandatory, remove it

Romainhir (approved):
Nice cosmetic feature ! Small questions in comment but otherwise LGTM

app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
armouldr: Yes you're right haha I forgot to do it, it's done :)

app/src/main/java/com/android/unio/ui/home/Home.kt
```diff
@@ -102,152 +90,199 @@ fun HomeScreen(
...
-                                    }
-                                    .padding(horizontal = 16.dp)
-                                    .onGloballyPositioned { coordinates ->
-                                      followingTabWidth = (coordinates.size.width / density).dp
-                                      followingTabXCoordinate = coordinates.positionInRoot().x
-                                    }
-                                    .testTag(HomeTestTags.TAB_SAVED))
-                      }
+@Composable
+fun HomeContent(
```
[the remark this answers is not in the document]
armouldr: After discussing this IRL, we concluded that it would cause problems to do this so I will leave the View models for now. Same for the other comments saying to use hilt.

app/src/main/java/com/android/unio/ui/home/Home.kt
armouldr: I don't think the constant being here is bad practice, although I am not 100% sure of it. I'll leave it for now and we can see later what we would do with this kind of constants

app/src/main/java/com/android/unio/ui/home/Home.kt
armouldr: The height is not hardcoded as sizeList is initalised by reading the tabs' width and height : `Modifier.testTag(tabTestTags[index]).onSizeChanged {
                          sizeList[index] = Pair(it.width.toFloat(), it.height.toFloat())
                        }`
I verified this by trying to launch the app on a virtual tablet, and the indicator bar seems to adapt well

app/src/main/java/com/android/unio/ui/home/Home.kt
armouldr: Yes, I just tried, onSearch is mandatory sadly

app/src/main/java/com/android/unio/ui/home/Home.kt
armouldr: same, it is mandatory :/

armouldr:
I am merging as the CI is temporarily bugged and the code had the required code coverage before some minor fixes.

PR #170, opened 2024-11-14
Refactor/repositories firestore operations

Zafouche:
## Description

This PR extracts the `performFirestoreOperation` function from a couple of repositories and consolidates it into a utility file, named `FirestoreUtils`. This function, which handles `Firestore` operations by invoking success and failure callbacks instead of the previous simple `onCompleteListener`, was either used in some repositories or not at all. The function was also transformed into an extension function, to simplify its usage.

## Motivation and context
The goal of this refactor is to improve code maintainability and readability, and to ensure callbacks are handled consistently.

### FirestoreUtils:
- Created this new utility file for Firestore utility functions.
- The `performFirestoreOperation` function has been transformed into an extension function on Task<T>
- Added error handling for cases where the result might be null, even though the `onSuccess` callback is invoked.

### Refactoring of Repositories:
- Updated `AssociationRepositoryFirestore`, `EventRepositoryFirestore`, `ImageRepositoryFirebaseStorage`, and `UserRepositoryFirestore` to use the `performFirestoreOperation` extension function.
- Removed the duplicate implementations of the `performFirestoreOperation` function from these repositories.

### Tests:
- Refactored tests in `AssociationRepositoryFirestoreTest` and `EventRepositoryFirestoreTest` to account for the change in the `performFirestoreOperation` implementation.

## How has this been tested?
The existing tests check that the repositories behave as expected.
I also tested manually, the app is fully functional with this refactor.

### DIFF
--- a/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
@@ -1,13 +1,14 @@
 package com.android.unio.model.association
 
 import com.android.unio.model.firestore.FirestorePaths.ASSOCIATION_PATH
+import com.android.unio.model.firestore.performFirestoreOperation
 import com.android.unio.model.firestore.transform.hydrate
 import com.android.unio.model.firestore.transform.serialize
-import com.google.android.gms.tasks.Task
 import com.google.firebase.Firebase
 import com.google.firebase.auth.auth
 import com.google.firebase.firestore.DocumentReference
 import com.google.firebase.firestore.FirebaseFirestore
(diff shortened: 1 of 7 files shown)


Zafouche:
Following the coverage analysis fix, which I just pulled from main, some minor changes have been made after them.

The `performFirestoreOperation` function has been slightly adjusted in the following way for the onSuccess listener of the function: 
- A write operation makes it such that the result is of type Void (so it can be null), meaning onSuccess should be called.
- A read operation makes it such that the result is of type DocumentSnapshot, and the null documents should be properly handled (logged)

This ensures that write operations (which can return null if successful) still work, e.g. creating an account, and that read operations are handled as supposed to.
Opening this for review.

PR #171, opened 2024-11-14
Fix CI not uploading coverage

oskar-codes:
This is an urgent fix needed as it appears that since #155 was merged, the coverage of Android tests isn't uploaded to SonarCloud.


oskar-codes:
Closed this branch in order to rename it to `branch-fixcoverage` to make it a long-lasting branch, to better test with sonarcloud.

PR #174, opened 2024-11-14
Fix CI not uploading coverage: branch-fixci

Zafouche:
This is another PR to try to address the problem mentioned in #172.


oskar-codes (approved):
Much needed fix and we finally did it (after too many hours but still) 💪

PR #175, opened 2024-11-15
Add [Test search feature] New end to end test for search feature

AlouchLaBouche:
This PR adds a new end to end test, testing the full functionality of the search feature done by @Redd87. This includes the association search and event search. 

This PR includes : 
- 2 new tests, testing the behavior of the app when searching for an association and for an event.
- A lot of code refactoring and extraction, especially regarding the end to end tests using the firebase emulator. More precisely I added an Interface for all tests that require the emulator, this is too improve code readability and decrease code duplication.
- Some other chores such as adding new test tags to necessary fields.

**Warning** This test only works if you have launched the emulators with populated data inside them. This data is found in `firebase/emulator-data` directory (this is the unreadable binary files that has to be pushed). To do so simply run : `firebase emulators:start --import = firebase/emulator-data`. When done so you'll see in the emulator UI, there is an preexisting data for some mock association, users and events.
When done so the tests should pass.


oskar-codes (approved):
Solid work on this, I'll approve this as I see no flaws in the resulting code.

PR #179, opened 2024-11-18
Link followed event to the frontend

Romainhir:
Currently, a user can follow an association, but cannot see their events in the following page, on the home screen. This is the purpose of this task. It aims to display the event from the followed association. The follow mechanism is already implemented so it just need to get the event and display it. Precisely, this means :

- Get the events from all followed association
- Display it on the follow screen
- Update accordingly on follow/unfollow


app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
```diff
@@ -98,15 +98,17 @@ open class EndToEndTest : FirebaseEmulatorFunctions {
...
     const val PORT = 8080
     const val ROOT = "http://$HOST:$PORT"
     const val SHORT_ROOT = "$HOST:$PORT"
     const val DATABASE_URL = "$ROOT/emulator/v1/projects/unio-1b8ee/databases/(default)/documents"
   }
 
   object Auth {
-    const val HOST = "10.0.2.2"
+//    const val HOST = "10.0.2.2"
+    const val HOST = "128.179.201.248"
```
Zafouche: Don't forget to undo these changes, as they are specific to your case (e.g. running the tests on your phone)

app/src/androidTest/java/com/android/unio/end2end/UserAccountCreationTest.kt
```diff
@@ -140,7 +140,7 @@ class UserAccountCreationTest : EndToEndTest() {
   private fun verifyEmail(url: String) {
     val client = OkHttpClient()
 
-    val request = Request.Builder().url(url.replace("127.0.0.1", "10.0.2.2")).build()
+    val request = Request.Builder().url(url.replace("127.0.0.1", "128.179.201.248")).build()
```
Zafouche, line 143: Same as above.

app/src/androidTest/java/com/android/unio/ui/home/HomeTest.kt
Zafouche, line 1: Good additional tests.

firebase.json
```diff
@@ -8,20 +8,33 @@
...
       "port": 9099
     },
     "firestore": {
-      "host": "127.0.0.1",
+      "host": "0.0.0.0",
       "port": 8080
     },
     "storage": {
-      "host": "127.0.0.1",
+      "host": "0.0.0.0",
```
Zafouche, line 19: I assume these are also specific to your testing setup?

Zafouche (commented):
Overall the code looks good.

However, the feature does not work on my end. I have tried to launch the app, login with my account and save an event, and unfortunately it does not show up on my `Following` tab of the `Home` screen, so I cannot approve this PR yet.

Does it work on your end? 

Also, please don't forget to undo your changes to the IPs used for the e2e tests, as they are specific to the case of running the tests on your phone.

Romainhir:
> save an event

Actually, I did not implement "saved event" but the event of the following association. Otherwise thank you for comments !

Zafouche (approved):
> > save an event
> 
> Actually, I did not implement "saved event" but the event of the following association. Otherwise thank you for comments !

I apologize, I confused the PR title and the branch name with the feature that displays the event we liked, but that shows up in the `Saved` screen, and not under the `Following` tab of the `Home` screen, which displays events for associations we follow.

In that case, the feature works perfectly fine and I have no bugs to report. Good job on this! However don't forget to change the IPs back because the e2e files do not pass on machines that use the emulator instead of *your* phone.

PR #180, opened 2024-11-18
Minor end-to-end testing utilities

oskar-codes:
This pull request adds some minor utilities that aim to help running end-to-end tests locally. These are only helpful tools for developers and are not meant to be included in the CI. There are two Gradle commands that have been added:
- `startFirebaseEmulators` Simply starts the Firebase emulators as defined by `firebase.json` and imports the data stored in `firebase/emulator-data`.
- `connectedCheckWithFirebaseEmulators` Same as above, but also runs `connectedCheck` when the emulators are launched and closes the emulators after the tests are executed.

Note that I have only tested these Gradle commands on Windows as I only have a Windows PC. Please let me know if they do not work on Linux or MacOS.

This pull request also cleans up a little the `EndToEndTest` class, with better error handling and constant extraction.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -366,6 +366,17 @@ fun runCommand(command: String) {
                 konsole -e '$command'
                 """.trimIndent()
         ).start()
+    } else if (os.contains("mac")) {
+        ProcessBuilder(
+            "osascript",
+            "-e",
+            """
+                tell application "Terminal"
+                    do script "firebase emulators:start --import=./firebase/emulator-data"
+                    activate
+                end tell
+                """.trimIndent()
+        ).start()
     } else {
         throw GradleException("Unsupported operating system: $os")
     }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -361,9 +361,9 @@ fun runCommand(command: String) {
             "sh",
             "-c",
             """
-                x-terminal-emulator -e '${command}' ||
-                gnome-terminal -- bash -c '${command}; exec bash' ||
-                konsole -e '${command}'
+                x-terminal-emulator -e '$command' ||
+                gnome-terminal -- bash -c '$command; exec bash' ||
+                konsole -e '$command'
                 """.trimIndent()
         ).start()
     } else {
@@ -374,5 +374,11 @@ fun runCommand(command: String) {
 tasks.register("startFirebaseEmulators") {
     doLast {
         runCommand("firebase emulators:start --import=./firebase/emulator-data")
+    }
+}
+
+tasks.register("connectedCheckWithFirebaseEmulators") {
+    doLast {
+        runCommand("firebase emulators:exec --import=./firebase/emulator-data \"gradlew connectedCheck\"")
     }
 }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -355,15 +355,25 @@ configurations.forEach { configuration ->
 fun runCommand(command: String) {
     val os = System.getProperty("os.name").toLowerCase()
     if (os.contains("win")) {
-        ProcessBuilder("cmd", "/c", "start", "cmd", "/k", command).start()
+        val adjustedCommand = if (command.contains("firebase emulators:exec")) {
+            command.replace("./gradlew", "gradlew") // remove ./ for Windows
+        } else {
+            command // other commands
+        }
+        ProcessBuilder("cmd", "/c", "start", "cmd", "/k", adjustedCommand).start()
     } else if (os.contains("nix") || os.contains("nux") || os.contains("bsd")) {
+        val adjustedCommand = if (command.contains("firebase emulators:exec")) {
+            command.replace("./gradlew", "gradlew") // remove ./ for Linux
+        } else {
+            command // other commands
+        }
         ProcessBuilder(
             "sh",
             "-c",
             """
-                x-terminal-emulator -e '$command' ||
-                gnome-terminal -- bash -c '$command; exec bash' ||
-                konsole -e '$command'
+                x-terminal-emulator -e '$adjustedCommand' ||
+                gnome-terminal -- bash -c '$adjustedCommand; exec bash' ||
+                konsole -e '$adjustedCommand'
                 """.trimIndent()
         ).start()
     } else if (os.contains("mac")) {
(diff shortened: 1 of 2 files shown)


app/build.gradle.kts
```diff
@@ -352,24 +352,44 @@ configurations.forEach { configuration ->
...
+            "-c",
+            """
+                x-terminal-emulator -e '$command' ||
+                gnome-terminal -- bash -c '$command; exec bash' ||
+                konsole -e '$command'
+                """.trimIndent()
+        ).start()
+    } else if (os.contains("mac")) {
+        ProcessBuilder(
+            "osascript",
```
Aurelien9Code, line 371: Thanks for using osascript to add support for macOS ; )

app/build.gradle.kts
```diff
@@ -352,24 +352,44 @@ configurations.forEach { configuration ->
...
+    } else if (os.contains("nix") || os.contains("nux") || os.contains("bsd")) {
+        ProcessBuilder(
+            "sh",
+            "-c",
+            """
+                x-terminal-emulator -e '$command' ||
+                gnome-terminal -- bash -c '$command; exec bash' ||
+                konsole -e '$command'
+                """.trimIndent()
+        ).start()
```
Aurelien9Code, line 368: As you can see at the bottom of the [Java documentation for ProcessBuilder.start()](https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/ProcessBuilder.html), the method can throw an `IOException`: "Starting an operating system process is highly system-dependent. Among the many things that can go wrong are:

    The operating system program file was not found.
    Access to the program file was denied.
    The working directory does not exist. 

In such cases an exception will be thrown. The exact nature of the exception is system-dependent, but it will always be a subclass of [IOException](https://docs.oracle.com/javase/8/docs/api/java/io/IOException.html).". Consider handling it with a `catch (e: IOException)`, and other errors with `catch (e: Exception)`.

app/build.gradle.kts
```diff
@@ -352,24 +352,44 @@ configurations.forEach { configuration ->
   configuration.exclude("com.google.protobuf", "protobuf-lite")
 }
 
+fun runCommand(command: String) {
+    val os = System.getProperty("os.name").toLowerCase()
+    if (os.contains("win")) {
```
Aurelien9Code, line 357: Could be replaced with `when`conditional logic, as you already have three cases, and as it is better for fallbacks.

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
```diff
@@ -62,19 +62,17 @@ open class EndToEndTest : FirebaseEmulatorFunctions {
...
-                assert(data.contains("Ok")) { "Firebase Emulators are not running." }
               }
             })
   }
 
   override fun useEmulators() {
     try {
-      Firebase.firestore.useEmulator(Firestore.HOST, 8080)
-      Firebase.auth.useEmulator(Auth.HOST, 9099)
+      Firebase.firestore.useEmulator(HOST, Firestore.PORT)
```
Aurelien9Code, line 74: Well done ! Seems really more readable now : )

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
```diff
@@ -62,19 +62,17 @@ open class EndToEndTest : FirebaseEmulatorFunctions {
               }
 
               override fun onResponse(call: okhttp3.Call, response: okhttp3.Response) {
-                if (response.body == null) {
+                if (response.body == null || !response.body!!.string().contains("Ok")) {
```
Aurelien9Code, line 65: Haha I have no idea how to change that but isn't there is a Json response ? That means we could check that the response is OK without consuming the response body. If not that seems really fine

Aurelien9Code (approved):
Great job ! Well done for creating the runCommand function ! It's making the OS-specific logic reusable across multiple tasks ; )
Good job to check if the response body is OK.

Maybe handle more some possible exceptions as explained but overall it's perfect !  🚀

app/build.gradle.kts
oskar-codes: Indeed, that's intentional :)
If an error occurs when running the command, it should be shown to the developper running it.

app/build.gradle.kts
```diff
@@ -352,24 +352,44 @@ configurations.forEach { configuration ->
...
+                tell application "Terminal"
+                    do script "$command"
+                    activate
+                end tell
+                """.trimIndent()
+        ).start()
+    } else {
+        throw GradleException("Unsupported operating system: $os")
+    }
+}
```
Zafouche, line 383: It does not work unfortunately, it runs the script in 200ms, and although returning a successful result, it obviously does not run the tests.

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
oskar-codes: The body isn't a JSON response, it is is simply the string `Ok` when the emulator is successfully running. Thus, it should be okay to consume the body as it will not be used anywhere else.

Aurelien9Code (changes requested):
OK, in the end we tested on Linux with @armouldr  ! 
Now we know that it works. 
HOWEVER : everytime we restart the mac or the PC using linux, we need to run `chmod +x gradlew` in the root directory of the project. There wouldn't be so much of sense to this PR if instead of running a certain command we need to run a certain other - but only if you have a specfic OS. 
So I would advise just to do a chmod in root directory for linux and macos, before runing connectedCheck.

oskar-codes:
I believe the issue regarding execute permissions on `gradlew` is caused whenever you pull the file from Github, as it does not have the required permissions. We should consider pushing a `gradlew` file after granting execute permissions on it. 

Other than that, I have cleaned up the code and fixed an issue occurring on Windows, let me know when I can merge @Aurelien9Code.

Aurelien9Code:
> I believe the issue regarding execute permissions on `gradlew` is caused whenever you pull the file from Github, as it does not have the required permissions. We should consider pushing a `gradlew` file after granting execute permissions on it.
> 
> Other than that, I have cleaned up the code and fixed an issue occurring on Windows, let me know when I can merge @Aurelien9Code.

Okay I see ! I will push the gradlew updated with chmod access in my PR ; )

PR #182, opened 2024-11-18
Feat/map gps refresh

Zafouche:
## Description

This PR implements periodic location updates to the `MapViewModel` using existing `FusedLocationProviderClient`. The periodic updates allow the app to refresh the location displayed on the map at regular intervals between 5-10 seconds. The logic ensures the GPS shows live location updates, such that the user can see their movements in real time on the map. The time interval can be discussed and adapted later, depending on the load of the rest of the app on a phone, but 5 seconds seems like a fair interval for our map.

## Motivation and Context

The `FusedLocationProviderClient` provides automatic location updates, which had to be coupled with the `MapViewModel` to be able to use it in our `Map` screen. This feature ensures the map reflects the user’s live current location while moving.

### MapViewModel:
- Added `startLocationUpdates` to initialize periodic location fetching with intervals.
- Added `stopLocationUpdates` to stop location updates when not needed (e.g., when leaving the map screen).
- Integrated LocationRequest for specifying update intervals (10 seconds) and the fastest update interval (5 seconds, which is the interval used pretty much 90% of the time by the `FusedLocationProviderClient`). It uses a LocationCallback, which has been overrode to update the `_userLocation` `MutableStateFlow` of the `MapViewModel`.
- Properly handles permissions to ensure location updates start only when required permissions are granted.

### MapScreen:
- Changed MapViewModel calls to start automatic location requests and end the periodic requests when the screen is exited.
- Periodic requests are made and update the user's location on the map.

## How Has This Been Tested?
- Added unit tests for `startLocationUpdates` and `stopLocationUpdates` in `MapViewModelTest`. All use cases of the functions have been tested.
- Verified location updates trigger correctly under various conditions (e.g., with permissions granted or denied).
- Tested that the location is updated correctly in the state flow.


Romainhir (approved):
LGTM, nice job !

PR #184, opened 2024-11-18
Feature/EventDetails: Link Event Details to map

armouldr:


armouldr:
Mistake when creating the PR, I did a new one with only my changes and I'm locking this one.

PR #186, opened 2024-11-18
Link EventDetails to the Map Screen

armouldr:
Link the map button in EventDetails to the map screen, such that the map is centered to the event in question.

For that I need to :
-> add a centerLocation stateflow in MapViewModel
-> add a setCenterLocation method in MapViewModel
-> modify the map screen accordingly

How it works:
in Home: simply navigate to Map without specifying any location, this will focus on userLocation
in EventDetails : use setCenterLocation method to

### DIFF
--- a/app/src/main/java/com/android/unio/MainActivity.kt
+++ b/app/src/main/java/com/android/unio/MainActivity.kt
@@ -47,8 +47,8 @@ import com.android.unio.ui.theme.AppTheme
 import com.android.unio.ui.user.UserProfileScreen
 import dagger.hilt.android.AndroidEntryPoint
 import dagger.hilt.android.HiltAndroidApp
-import me.zhanghai.compose.preference.ProvidePreferenceLocals
 import javax.inject.Inject
+import me.zhanghai.compose.preference.ProvidePreferenceLocals
 
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
@@ -116,7 +116,7 @@ fun UnioApp(imageRepository: ImageRepositoryFirebaseStorage) {
             navigationAction = navigationActions,
             eventViewModel = eventViewModel,
             userViewModel = userViewModel,
-          mapViewModel)
+            mapViewModel)
       }
       composable(Screen.MAP) {
         MapScreen(navigationActions, eventViewModel, userViewModel, mapViewModel)
--- a/app/src/main/java/com/android/unio/MainActivity.kt
+++ b/app/src/main/java/com/android/unio/MainActivity.kt
@@ -47,8 +47,8 @@ import com.android.unio.ui.theme.AppTheme
 import com.android.unio.ui.user.UserProfileScreen
 import dagger.hilt.android.AndroidEntryPoint
 import dagger.hilt.android.HiltAndroidApp
-import javax.inject.Inject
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/android/unio/MainActivity.kt
```diff
@@ -115,7 +115,8 @@ fun UnioApp(imageRepository: ImageRepositoryFirebaseStorage) {
         EventScreen(
             navigationAction = navigationActions,
             eventViewModel = eventViewModel,
-            userViewModel = userViewModel)
+            userViewModel = userViewModel,
+            mapViewModel)
```
Zafouche, line 119: Keep the way you input the parameters consistent, e.g. `mapViewModel = mapViewModel`

app/src/main/java/com/android/unio/ui/map/Map.kt
```diff
@@ -121,16 +122,20 @@ fun MapScreen(
...
       }
     }
-    cameraPositionState.position =
-        CameraPosition.fromLatLngZoom(userLocation ?: EPFL_COORDINATES, INITIAL_ZOOM_LEVEL)
   }
 
-  // Center map on the user's location initially if available
+  // Center map on the center location initially if available
   LaunchedEffect(userLocation) {
-    if (userLocation != null && !initialCentered) {
```
Zafouche, line 130: I don't think removing this boolean is a good idea, as if the `userLocation` state flow is updated, e.g. the user moves to a different location and their location is updated in the view model, the map would automatically recenter on one of the logics below (either centers on the user's location, event's location or on EPFL's location) for no reason, and that's a behavior we don't want.

Zafouche (changes requested):
Overall good work, when we click on the address of the event in its respective page, it indeed brings us to the event's location. Great feature!

However a bug is introduced, so I cannot approve this PR yet. Have a look at my second comment. I know automatic location refresh isn't present on the `main` branch yet, but it is about to be through PR #182. Event if we had manually written location updates (calling `mapViewModel.fetchUserLocation`) occasionally, this wouldn't work. I know you can't really walk around and change your GPS location in the emulator, but don't forget the case where the user can actually walk around!

Otherwise, this feature is really neat and useful so thanks!

Zafouche (approved):
This now LGTM.

While playing around with location permissions and opening the map, I have found a small issue: Say it is the first time you open the map, and you refuse to share your location. The map centers on EPFL's coordinates. And say you manually grant permission for the app to use the phone's location. If you re-open the map after that, it still centers on EPFL's coordinates, but that's not what we want. What we want, is for the map to center itself on the user. Essentially, this can easily be fixed by changing `intialCentered` to a non-mutable state, e.g. something that does back to false whenever we come back to the screen. This would make more sense. And also, having the user's location being refreshed often would help re-triggering the launched effect that handles the re-centering, which depending on when the FusedLocationProviderClient gives a location, can slightly delay the re-center. But as you know, this will be coming in the PR #182.

This small bug is my fault, it wasn't noticeable with the first map draft I had implemented, sorry!

Zafouche:
@armouldr, I invite you to change `intialCentered` from `var initialCentered by remember { mutableStateOf(false) }` to `var initialCentered = false` before merging. This fixes the undesired behavior I described above.

PR #187, opened 2024-11-19
[Refactor AccountDetails] Refactor of the account details composable 

AlouchLaBouche:
This small PR is a well deserved refactor of the `AccountDetails` Composable. 
No new functionality has been added, the only changes are better uses of callbacks functions to extract the viewModels from the main composable (this is to enable `@Previews`), and hopefully a more readable code :smile: .

I also refactored the social and interests overlays!


app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
```diff
@@ -70,13 +71,57 @@ import com.google.firebase.Firebase
...
+      navigationAction,
+      onCreateUser = { profilePictureUri, createUser ->
+        if (profilePictureUri.value == Uri.EMPTY) {
+          createUser("", userId!!)
+        } else {
+          val inputStream = context.contentResolver.openInputStream(profilePictureUri.value)
+          imageRepository.uploadImage(
+              inputStream!!,
+              "images/users/${userId}",
+              onSuccess = { createUser("images/users/${userId}", userId!!) },
```
armouldr, line 94: This hardcoded Firebase storage path could be moved to Strings.kt along with the Firestore ones

app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
```diff
@@ -158,174 +185,33 @@ fun AccountDetails(
...
-                      })
-                }
-              }) {
+              /**
+               * The [onCreateUser] function is called with the [profilePictureUri] and [createUser]
+               * lambda function as parameters. [createUser] calls upon another lambda that uploads
+               * the user if all the required fields are filled. This method hierarchy is necessary
+               * due to the fact that uploading an image needs to know if the URI is empty or not
+               * before creating the User.
+               */
```
armouldr, line 213: I don't think it's good practice to put javadoc in the body of a function, you could make it a simple comment or you could create a javadoc for AccountDetailsContent and put this explanation in it. But it's a good initiative to explain in detail this pretty complex process :)

armouldr (commented):
A nice refactor that indeed makes the screen's code more readable, everything still works as expected. Nice work :)

app/src/androidTest/java/com/android/unio/end2end/UserAccountCreationTest.kt
```diff
@@ -39,21 +39,23 @@ class UserAccountCreationTest : EndToEndTest() {
...
     /** Refresh the email verification and continue */
-    composeTestRule.onNodeWithTag(EmailVerificationTestTags.SCREEN).assertIsDisplayed()
+    composeTestRule.waitUntil(10000) {
+      composeTestRule.onNodeWithTag(EmailVerificationTestTags.SCREEN).isDisplayed()
+    }
     composeTestRule.onNodeWithTag(EmailVerificationTestTags.REFRESH).performClick()
 
-    Thread.sleep(5000)
-
+    composeTestRule.waitUntil(100000) {
```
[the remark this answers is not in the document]
AlouchLaBouche: Ah yes I wanted to put 10'000, and not 100'000, that's a mistake on my part, well spotted.

app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
AlouchLaBouche: Well noticed, I'll do just that :+1:

app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
AlouchLaBouche: You're right but because it is a lambda and not a "function" so to say, I didn't really know where to put it

armouldr (approved):
Thanks for the changes @AlouchLaBouche  ! LGTM, just make sure to correct the typo before merging :rocket:

PR #189, opened 2024-11-19
Implement start and end date for events

armouldr:
As of now we only had a single date for events, e.g the event's start date. But this isn't enough information as we would like to know the event's duration, as well as inform the users of the hour/ day on which the event ends.

So this PR includes:
-> refactor of the Event data class to include startDate and endDate.
-> Change EventDetails.kt accordingly so that it displays these informations in a nice manner
-> Maybe also modify EventCard (TO BE DEBATED: do we need the endDate on EventCard ?)

### DIFF
--- a/app/src/main/java/com/android/unio/mocks/event/MockEvent.kt
+++ b/app/src/main/java/com/android/unio/mocks/event/MockEvent.kt
@@ -107,7 +107,7 @@ class MockEvent {
           description = description,
           catchyDescription = catchyDescription,
           price = price,
-          date = date,
+          startDate = date,
           location = location,
           types = types,
           placesRemaining = -1)
--- a/app/src/main/java/com/android/unio/ui/event/EventCard.kt
+++ b/app/src/main/java/com/android/unio/ui/event/EventCard.kt
@@ -230,7 +230,9 @@ fun EventCardScaffold(
                 modifier =
                     Modifier.padding(vertical = 1.dp, horizontal = 0.dp)
                         .testTag(EventCardTestTags.EVENT_DATE),
-                text = formatTimestamp(event.date, SimpleDateFormat("dd/MM", Locale.getDefault())),
+                text =
+                    formatTimestamp(
+                        event.startDate, SimpleDateFormat("dd/MM", Locale.getDefault())),
                 style = MaterialTheme.typography.bodySmall,
                 color = MaterialTheme.colorScheme.onSurface)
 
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/android/unio/ui/event/EventCard.kt
```diff
@@ -230,7 +230,9 @@ fun EventCardScaffold(
                 modifier =
                     Modifier.padding(vertical = 1.dp, horizontal = 0.dp)
                         .testTag(EventCardTestTags.EVENT_DATE),
-                text = formatTimestamp(event.date, SimpleDateFormat("dd/MM", Locale.getDefault())),
+                text =
+                    formatTimestamp(
+                        event.startDate, SimpleDateFormat("dd/MM", Locale.getDefault())),
```
Redd87, line 235: The SimpleDateFormat("dd/MM", Locale.getDefault())) should really be extracted to a constant in case we ever want to change the date format at some point since it's used in ~6 different places. Do it now if you have time or just add it as a task for the future since it's not urgent

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
```diff
@@ -60,7 +60,7 @@ class HydrationAndSerializationTest {
           description = "An example event",
           catchyDescription = "An example event",
           price = 0.0,
-          date = Timestamp.now(),
+          startDate = Timestamp.now(),
```
Redd87, line 63: The events here should also include an end date

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
```diff
@@ -148,7 +149,7 @@ class HydrationAndSerializationTest {
     assertEquals(event.description, hydrated.description)
     assertEquals(event.catchyDescription, hydrated.catchyDescription)
     assertEquals(event.price, hydrated.price)
-    assertEquals(event.date, hydrated.date)
+    assertEquals(event.startDate, hydrated.startDate)
```
Redd87, line 152: Along with the corresponding assert for the end date

Redd87 (changes requested):
Good job overall! Just fix these small changes and I'll approve the PR it shouldn't take long normally

app/src/main/java/com/android/unio/mocks/event/MockEvent.kt
```diff
@@ -107,7 +107,7 @@ class MockEvent {
           description = description,
           catchyDescription = catchyDescription,
           price = price,
-          date = date,
+          startDate = date,
```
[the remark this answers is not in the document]
armouldr: Yes you're right I forgot to update the Mock and tests with this new change !

app/src/main/java/com/android/unio/ui/event/EventCard.kt
armouldr: As SimpleDateFormat does not work well with being in a static context, I decided to just put the formatting strings in Strings.kt :)

PR #195, opened 2024-11-20
Fix: Home page navigation

armouldr:
There is a bug with the Top bar navigation that makes the Screen display 2 times the same screen during the scroll. I fixed it by using HorizontalPager correctly (using its page value-parameter instead of pagerState.currentPage).


Romainhir (approved):
LGTM, nice and elegant fix !

PR #197, opened 2024-11-20
Feature/saved events v2

Redd87:
This PR continues the PR started by @Aurelien9Code some time ago and populates the saved events with events from the backend according to the Figma design. The screen is quite simple and shows the events that are happening today preceded by a delimiter, if any, and does the same for events that are not today. If no events are saved it shows "No events saved".

The whole thing is tested with testTags on the key elements.


app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    field.invoke(userViewModel, listOf(asso.uid))
+
+    eventList =
+        listOf(
+            MockEvent.createMockEvent(organisers = listOf(asso)),
+            MockEvent.createMockEvent(title = "I am different", startDate = Timestamp.now()))
+
+    val eventField = eventViewModel.javaClass.getDeclaredMethod("setEvents", List::class.java)
+    eventField.isAccessible = true
+    eventField.invoke(eventViewModel, eventList)
```
armouldr, line 98: Here I get it more easily, as we made it harder to add events to eventViewModel, but I think we will have to refactor EventVIewModel a bit in a next PR

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    every { userViewModel.isEventSavedForCurrentUser(any()) } returns true
+
+    assert(eventViewModel.events.value.isNotEmpty())
```
armouldr, line 102: It doesn't seem very logical to me to have an assert in a setUp method

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
```diff
@@ -0,0 +1,130 @@
...
+    eventListFollowed = asso.let { eventList.filter { event -> event.organisers.contains(it.uid) } }
```
armouldr, line 103: I don't think this variable is used anywhere, tell me if I'm wrong

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
```diff
@@ -0,0 +1,130 @@
...
+  @MockK(relaxed = true) private lateinit var searchRepository: SearchRepository
+
+  private lateinit var eventList: List<Event>
+  private lateinit var eventListFollowed: List<Event>
+
+  @Before
+  fun setUp() {
+    MockKAnnotations.init(this)
+    hiltRule.inject()
+    searchViewModel = spyk(SearchViewModel(searchRepository))
```
armouldr, line 62: This variable doesn't seem to be used anywhere

app/src/main/java/com/android/unio/ui/saved/Saved.kt
```diff
@@ -1,28 +1,128 @@
...
+  val savedEvents = allEvents.filter { userViewModel.isEventSavedForCurrentUser(it.uid) }
   val context = LocalContext.current
+
+  val today = remember { Calendar.getInstance().get(Calendar.DAY_OF_YEAR) }
+  val savedEventsToday = remember {
+    savedEvents.partition {
+      val calendar = Calendar.getInstance()
+      calendar.timeInMillis = it.startDate.toDate().time
+      calendar.get(Calendar.DAY_OF_YEAR) == today
+    }
```
armouldr, line 54: In my opinion it would be better to divide this pair into 2 variables : savedEventsToday and savedEventsUpcoming so the name is less confusing. The UI separation between today and upcoming is a good idea though and it is nicely coded !

app/src/main/res/values-fr/strings.xml
```diff
@@ -200,6 +200,13 @@
     * package: saved
     */
 
+    <!-- Saved Screen Strings -->
+    <string name="saved_screen_title">Evènements sauvegardés</string>
+    <string name="saved_screen_no_events">Aucun évènements sauvegardé</string>
+    <string name="saved_screen_today">Aujourd\'hui</string>
+    <string name="saved_screen_upcoming">Bientôt</string>
+
```
armouldr, line 208: It's great that you thought about translating the strings :+1:

armouldr (changes requested):
A nice PR overall, but some small adjustements need to be made, especially to the test class. But the new screen works well and is a much needed addition to the project :rocket:

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
Redd87: I agree ; I just copied the setup from Home and forgot to trim the excess. No idea why we have an assert in the setup in home though

app/src/main/java/com/android/unio/ui/saved/Saved.kt
Redd87: And I agree that the name of savedEventsToday  could be changed to something like savedEventsPartitioned or at least something that better reflects its nature, but I like the solutin with the .partition and find it clean. It avoids code duplication and is more efficient

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
oskar-codes: Indeed, there is no reason to manually inject the event list when you have already mocked the repository's `getEvents` method. Additionally, tests shouldn't depend on the internal behavior of a view model, as it can be subject to change at any time. Consider removing this.

app/src/main/res/values-fr/strings.xml
```diff
@@ -200,6 +200,13 @@
     * package: saved
     */
 
+    <!-- Saved Screen Strings -->
+    <string name="saved_screen_title">Evènements sauvegardés</string>
+    <string name="saved_screen_no_events">Aucun évènements sauvegardé</string>
```
oskar-codes, line 205: Minor typo, `évènements` should be singular, e.g. `Aucun évènement sauvegardé`.

app/src/androidTest/java/com/android/unio/ui/event/EventCreationTest.kt
```diff
@@ -125,14 +122,3 @@ class EventCreationTest {
     assertDisplayComponentInScroll(composeTestRule.onNodeWithTag(EventCreationOverlayTestTags.SAVE))
   }
 }
-
-/**
- * This function is a copy of the function with the same name from EditAssociationTest.kt. It should
- * be extracted to a common file in a future PR.
- */
-private fun assertDisplayComponentInScroll(compose: SemanticsNodeInteraction) {
```
oskar-codes, line 133: This is not about your pull request, but we should remove all private `assertDisplayComponentInScroll` methods and use the one from the utils.

oskar-codes (commented):
Not much to say, but you should consider removing parts of your test that directly inject data into the view-models private fields, as that is not how a test should work.

Also in `Saved.kt`, consider directly accessing the saved events list instead of filtering all events by their existance in the saved events list.

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        {
+          val onSuccess = args[0] as (List<Event>) -> Unit
+          onSuccess(eventList)
+        }
+    eventViewModel = EventViewModel(eventRepository, imageRepository)
+
+    val field =
+        userViewModel.javaClass.getDeclaredMethod("setFollowedAssociations", List::class.java)
+    field.isAccessible = true
+    field.invoke(userViewModel, listOf(asso.uid))
```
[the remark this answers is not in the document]
Romainhir: Conceptually, this is not accurate. YOu want to plug your mock user in the private field, in case of any change made on it, it is reported applied and "reported" on the public field, used by the UI. So you have to set up the private field here liek this (you use a setter because directly change the value does not work for some reason).

app/src/androidTest/java/com/android/unio/ui/saved/SavedTest.kt
Romainhir: I do not agree on this. If you want to test the behavior of the view model while using the UI, this is needed

app/src/androidTest/java/com/android/unio/ui/event/EventCreationTest.kt
Redd87: You're right and this is done in my other PR connecting the event creation frontend to the backend

app/src/main/java/com/android/unio/ui/saved/Saved.kt
```diff
@@ -1,28 +1,128 @@
...
-@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
 @Composable
-fun SavedScreen(navigationAction: NavigationAction) {
+fun SavedScreen(
+    navigationAction: NavigationAction,
+    eventViewModel: EventViewModel,
+    userViewModel: UserViewModel
+) {
+  val allEvents by eventViewModel.events.collectAsState()
+  val savedEvents = allEvents.filter { userViewModel.isEventSavedForCurrentUser(it.uid) }
```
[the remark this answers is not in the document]
Redd87: I wasn't aware this existed, but it highly simplifies the code. Though I'm not sure where I should call .requestAll ? As you're the one who worked on the Userviewmodel, do you want me to modify getUserByUid to call savedEvents.requestAll ?

app/src/main/java/com/android/unio/ui/saved/Saved.kt
armouldr: You're right the partition is nice, I was only talking of then dividing the pair into 2 different lists

armouldr (approved):

PR #198, opened 2024-11-21
Feature/event creation backend

Redd87:
This PR aims to connect the event creation UI to the Firestore database. It modifies the EventCreation file to call eventViewModel.addEvent in the onClick of the create button and does all the required changes for it to work. It enforces a light check on the user arguments (ensures they are non-null and that the startTime < endTime). 
This is however being put as a draft for the moment as there are no tests at the moment and I'd like to test this with the firebase emulator.

### DIFF
--- a/app/src/main/java/com/android/unio/ui/event/EventCreation.kt
+++ b/app/src/main/java/com/android/unio/ui/event/EventCreation.kt
@@ -1,12 +1,10 @@
 package com.android.unio.ui.event
 
 import android.net.Uri
-import android.os.Build
 import android.widget.Toast
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.PickVisualMediaRequest
 import androidx.activity.result.contract.ActivityResultContracts
-import androidx.annotation.RequiresApi
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.awaitEachGesture
@@ -87,7 +85,6 @@ import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 
-@RequiresApi(Build.VERSION_CODES.TIRAMISU)
 @Composable
 fun EventCreationScreen(
     navigationAction: NavigationAction,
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -188,7 +220,48 @@ fun EventCreationScreen(
...
+                      startTimestamp != null &&
+                      endTimestamp != null &&
+                      startTimestamp!! < endTimestamp!! &&
+                      eventBannerUri.value != Uri.EMPTY,
+              onClick = {
+                val inputStream = context.contentResolver.openInputStream(eventBannerUri.value)!!
+                eventViewModel.addEvent(
+                    inputStream,
+                    Event(
+                        uid = "", // This gets overwritten by eventViewModel.addEvent
```
Zafouche, line 236: It would probably make more sense to create a `getNewUID` function in the `EventViewModel` that calls `repository.getNewUid()`, instead of letting the view model handle this implicitly, simply because it brings less confusion.

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -188,7 +220,48 @@ fun EventCreationScreen(
...
+                        taggedAssociations =
+                            Association.firestoreReferenceListWith(
+                                taggedAndBoolean.filter { it.second.value }.map { it.first.uid }),
+                        image = eventBannerUri.value.toString(),
+                        description = longDescription,
+                        catchyDescription = shortDescription,
+                        price = 0.0,
+                        startDate = startTimestamp!!,
+                        endDate = endTimestamp!!,
+                        location = Location(),
```
Zafouche, line 254: Just leaving this comment for future reference: This won't be an empty location when the Search Places API will be implemented (or Nominatim)

Zafouche (approved):
This LGTM, clean and efficient work! 🚀

Only one remark to make, but I gladly approve!

I trust you to not forget to add tests later on, coverage is important!

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
Redd87: I'm not sure the UI should interact directly with the repository? But let's discuss this tomorrow

PR #149, opened 2024-11-07
Refactor Association Profile and several other fixes

oskar-codes:
The main objective of this pull request is to fix an issue that came with #104, where a backend request was made directly from a Composable's scope in `AssociationProfile`, which thus made Firestore requests at every composable refresh.
This issue is now fixed, along with a refactor of the `AssociationViewModel`. To keep consistency accross the project, the same refactor was reflected in in `UserProfile`.
The most notable change of the refactor is to remove the use of the string parameter `{uid}` of the association profile route. Instead, the association is stored in the `AssociationViewModel`. This was decided after discussing with @armouldr, who is currently implementing the same process for the `EventDetails`.

PR #199, opened 2024-11-21
Refactor various files

oskar-codes:
This pull request aims to refactor several files that contain unnecessary excessive boilerplate code that increases cognitive complexity. Additionally, it adds more robust guard clauses to make our code stronger.

This pull request also fixes the issue introduced by #149, where users could not navigate to the associations they had joined or followed.

Additionally, it also updates all tests that bypass private field protection, as we should never rely on the internal behavior of classes when testing them.


Redd87 (commented):
Good job overall! Though I do have one question, should we remove the setters from the view models and let the _variable be modified directly in the code? I feel it is against convention to modify variables that start with a _ by bypassing the setter

oskar-codes:
> Good job overall! Though I do have one question, should we remove the setters from the view models and let the _variable be modified directly in the code? I feel it is against convention to modify variables that start with a _ by bypassing the setter

The variables prefixed with and underscore are private and thus only accessible from the class. This is intended and conventionally acceptable.

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
```diff
@@ -31,7 +31,7 @@ import kotlinx.coroutines.flow.StateFlow
  * @property collectionPath The path to the Firestore collection.
  * @property hydrate A function that converts a [DocumentSnapshot] to a [T].
  */
-class FirestoreReferenceList<T>(
+class FirestoreReferenceList<T : UniquelyIdentifiable>(
```
Zafouche, line 34: Interesting addition, could you please explain to me what value this brings?

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -260,11 +267,34 @@ private fun AssociationProfileContent(
...
+      enableButton = true
+    }
+    isFollowed = !isFollowed
+  }
+
+  val onMemberClick = { member: User ->
+    userViewModel.setSomeoneElseUser(member)
+    navigationAction.navigateTo(Screen.SOMEONE_ELSE_PROFILE)
+  }
+
```
Zafouche, line 297: I didn't realize most of the `UserViewModel` could be simply cleaned up to this, good catch.

Zafouche (approved):
LGTM. As always code improvement is much needed, great work!
I only have one question, other than that I gladly approve. 🚀

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
```diff
@@ -72,7 +72,16 @@ class FirestoreReferenceList<T>(
 
   /** Requests all documents from Firestore and updates the list. */
   override fun requestAll(onSuccess: () -> Unit) {
+    // If the list is already up-to-date, return early.
+    val fetchedUids = _list.value.map { it.uid }
+    if (fetchedUids == _uids) {
+      onSuccess()
+      return
+    }
```
Aurelien9Code, line 80: Very good idea ! However, if you keep this implementation, you won't fetch the database even if other components are modified remotely, given that the list of uids is the same. This is quite critical as it can happen in various context. 

As I am not far from you and I thus asked you this question, you proposed to add a boolean "forcedPush" which can bypass the request if we need to see changes from the database. I agree with you and this is a good way of handling the problem ! However I think we would like to see the updates from Firebase in almost every situation, given that we are in the Screen displaying them. The answer might be in [realtime updates with Cloud Firestore](https://firebase.google.com/docs/firestore/query-data/listen) -> 

We could have a ListenerManager that manage all the listeners to the database according to the current screen. If the current one is the Event list for example, with only 3 events visible on the page, we could say that there is one listener per event visible, and if we want to be meticulous, even only for the specific fields seen in the screen. Then if the user click on one event, we will kill the listeners of the two other events and add fields to the listener of the clicked event. This is just an idea, and of course outside the scope of this PR ; )

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -503,19 +526,14 @@ private fun AssociationDescription(association: Association) {
...
-    userViewModel: UserViewModel,
-    user: User
+    association: Association,
+    isFollowed: Boolean,
+    enableButton: Boolean,
+    onFollow: () -> Unit,
 ) {
-  val association = associationViewModel.selectedAssociation.collectAsState().value!!
-  var isFollowed by remember { mutableStateOf(user.isFollowAssociation(association)) }
-  var isButtonEnabled by remember { mutableStateOf(true) }
```
Aurelien9Code, line 518: Very good implementation to remove the view models ! We should now pay attention that the argument given to the @Composable like this are collectAsState thought, not to loose the automatic refresh ; )

app/src/main/java/com/android/unio/ui/event/EventCard.kt
```diff
@@ -73,7 +71,7 @@ fun EventCard(
     return
   }
 
-  var isSaved by remember { mutableStateOf(user!!.savedEvents.contains(event.uid)) }
```
Aurelien9Code, line 76: I see why you can remove the mutableState as it is always trigger by the onClick, good job !

Aurelien9Code (approved):
Very good job ! Looks perfect to me ; )
Check the big comment I made on the ReferenceList, as the current implementation is really a problem for the app.

Your CI is failing, however I see that it passed before and now it's just due to ktfmt format. I thus approve your PR.

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
Zafouche: Adding `Snapshot` listeners for sure requires our attention, I read about it while I was investigating offline mode.
 I don't understand quite well how this ListenerManager would work, but I'd be happy to talk about it tomorrow if you think it's worth it. 

However I've seen some practices of people defining an observer in their repositories, that listens to the database (that would also work locally, e.g. in offline mode) and update its contents (association list, event list,...). So I agree, `Snapshot` listeners are something we'll need to use, I think. We'll discuss it soon anyway.

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
oskar-codes: This enables the `FirestoreReferenceList` to access the `uid` field of elements in its `list` property. 
Thus, it makes it possible to do for example the check I added to `.requestAll` that avoids excessive requests.

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
oskar-codes: Indeed I agree with you both. For now I have added a `lazy` parameter to `.requestAll` which when set to true exits the function earlier. Its default value is false.

PR #202, opened 2024-11-21
refactor(access-rights): modify some files according to reviews

Aurelien9Code:
This PR fix a little bug we saw with @armouldr in the verification code.


AlouchLaBouche (approved):
Looks good to me :rocket: , well spotted for the bug fix :muscle:

PR #203, opened 2024-11-21
Feat/user edit settings

AlouchLaBouche:
This PR aims to add the ability for a user to edit his profile :rocket: . He can now, after creating his account, modify his first and last name, biography, interests, socials and profile picture. This is done through the `my profile` page and clicking on the top right icon, and then selecting Edit Profile.
This is extremely similar to the `account details` composable that was done a few sprints ago, and it was discussed whether to reuse it entirely or not for the `edit user profile`, but due to small differences that were difficult to parametrize such as test tags or differences in UI layout/texts, it was decided to add an entirely new composable. I tried my best to extract most of the components in the `UserEditionComponents.kt` file but there is still a lot of duplication due to reasons mentioned in the precious sentence.  
This PR is still a draft due to the merger of @Aurelien9Code branch, I'm experiencing some navigation issues, but is should be fixed pretty quickly!


app/src/androidTest/java/com/android/unio/end2end/EditUserDetailsTest.kt
Zafouche, line 1: Great e2e test!

app/src/androidTest/java/com/android/unio/end2end/UserAccountCreationTest.kt
```diff
@@ -72,20 +70,21 @@ class UserAccountCreationTest : EndToEndTest() {
     assertDisplayComponentInScroll(
         composeTestRule.onNodeWithTag(AccountDetailsTestTags.INTERESTS_BUTTON))
     composeTestRule.onNodeWithTag(AccountDetailsTestTags.INTERESTS_BUTTON).performClick()
+
     assertDisplayComponentInScroll(
-        composeTestRule.onNodeWithTag(InterestsOverlayTestTags.CLICKABLE_ROW + "0"))
+        composeTestRule.onNodeWithTag(InterestsOverlayTestTags.CLICKABLE_ROW + "SPORTS"))
```
Zafouche, line 75: Thanks for these refactors you've done across multiple files!

app/src/main/java/com/android/unio/ui/navigation/NavigationAction.kt
```diff
@@ -129,6 +129,7 @@ object Screen {
   const val EXPLORE = "Explore_Screen"
   const val SAVED = "Saved_Screen"
   const val MY_PROFILE = "MyProfile_Screen"
+  const val EDIT_PROFILE = "Edit_Profile_Screen"
   const val SOMEONE_ELSE_PROFILE = "SomeoneElseProfile_Screen"
   const val SETTINGS = "Settings"
   const val ASSOCIATION_PROFILE = "Association_Profile_Screen"
```
Zafouche, line 135: I just realized, here, we should settle on one format, e.g. `EditProfile_Screen` or `Edit_Profile_Screen`. If you look at the other screens, you can see we defined `SomeoneElseProfile_Screen`, and did not use an `_` for each word separation, only as a prefix for the `Screen` word. This is not your fault, but we should keep only one format.

app/src/main/java/com/android/unio/ui/user/UserProfileEdition.kt
```diff
@@ -0,0 +1,371 @@
...
+      onModifyUser = { profilePictureUri, createUser ->
+        val uriType = checkImageUri(profilePictureUri.value.toString())
+
+        when (uriType) {
+          ImageUriType.EMPTY -> createUser("") // create a user with an empty URI
+          ImageUriType.REMOTE ->
+              createUser(user!!.profilePicture) // create a user with the same URI
+          ImageUriType.LOCAL -> { // create a user with a new URI and upload the image to storage
+            val inputStream = context.contentResolver.openInputStream(profilePictureUri.value)
+            imageRepository.uploadImage(
```
Zafouche, line 85: I think it's odd to call a repository from the UI, we should probably add a view model for the image repository, to respect the MVVM model. This is an issue we should address in another PR (task), as it doesn't have much to do with your work.

Anywho, nice work on this screen! Thanks for explaining some of the ambiguous code in comments!

app/src/main/java/com/android/unio/ui/user/UserProfileEdition.kt
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  val userSocialsFlow = remember { MutableStateFlow(user.socials.toMutableList()) }
+
+  val interests by userInterestsFlow.collectAsState()
+  val socials by userSocialsFlow.collectAsState()
+
+  // This is the local uri of the new profile picture stored locally
+  // But if it's the first time entering the page the uri will be the one from firebase
+  // and therefor cannot be opened in a input stream in the onModifyUser function
```
Zafouche, line 142: Little typo on the word `therefor`

Zafouche (approved):
Exceptional work! 🚀

Really nothing to complain about on your PR. I've only found issues that are outside the scope of your work, so good job! We should address them in different PRs.

Great testing, line coverage remains strong 💪

app/src/main/java/com/android/unio/ui/user/UserProfileEdition.kt
AlouchLaBouche: Yes you are absolutely right, passing the repo directly to the view is definitely not the right way to go about this problem. I tried to implement the Repo directly into the `userViewModel` but it was a big refactoring job that was not in the scope of this task. 
But thanks for the heads up, we should definitely add a task for this job!

app/src/androidTest/java/com/android/unio/end2end/UserAccountCreationTest.kt
AlouchLaBouche: :+1:

Issue #208, opened 2024-11-22
Fix tests needing to pass in a certain order


oskar-codes:
This issue follows #177. Currently the tests pass, but only when ran in a specific order namely when the end to end tests are executed before all other tests. If they are placed at the end, they fail with a MockK stubbing error, indicating that an end to end test contains a `UserRepositoryFirestore` mock even though there shouldn't be any mocks. This is the error:
```
io.mockk.MockKException: no answer found for UserRepositoryFirestore(userRepository#255).init(lambda {}) among the configured answers: ()
at io.mockk.impl.stub.MockKStub.defaultAnswer(MockKStub.kt:91)
at io.mockk.impl.stub.MockKStub.answer(MockKStub.kt:42)
at io.mockk.impl.recording.states.AnsweringState.call(AnsweringState.kt:16)
at io.mockk.impl.recording.CommonCallRecorder.call(CommonCallRecorder.kt:53)
at io.mockk.impl.stub.MockKStub.handleInvocation(MockKStub.kt:269)
at io.mockk.impl.instantiation.JvmMockFactoryHelper$mockHandler$1.invocation(JvmMockFactoryHelper.kt:24)
at io.mockk.proxy.android.advice.Advice.handle$lambda$0(Advice.kt:78)
at io.mockk.proxy.android.advice.Advice.$r8$lambda$__tUSv4gdDmG-jSc9y9hgANlxW8(Unknown Source:0)
at io.mockk.proxy.android.advice.Advice$$ExternalSyntheticLambda0.call(D8$$SyntheticClass:0)
at com.android.unio.model.user.UserRepositoryFirestore.init(Unknown Source:40)
at com.android.unio.model.user.UserViewModel._init_$lambda$0(UserViewModel.kt:52)
at com.android.unio.model.user.UserViewModel.$r8$lambda$Z2bBUy5q0_Cmv_9uZD2LhCDRFuQ(Unknown Source:0)
at com.android.unio.model.user.UserViewModel$$ExternalSyntheticLambda1.onAuthStateChanged(D8$$SyntheticClass:0)
at com.google.firebase.auth.zzz.run(com.google.firebase:firebase-auth@@23.0.0:3)
at android.os.Handler.handleCallback(Handler.java:958)
at android.os.Handler.dispatchMessage(Handler.java:99)
at android.os.Looper.loopOnce(Looper.java:205)
at android.os.Looper.loop(Looper.java:294)
[...]

PR #209, opened 2024-11-24
Fix tests

oskar-codes:
This pull request fixes all issues we had regarding state persisting between tests and end to end tests failing when ran after other UI tests. If you wish to understand the root cause of the issue, the entire fix process has been documented [here](https://github.com/SwEnt-Group13/Unio/issues/208).

### 1. TearDown
When writing tests in the future, please use the `TearDown` class, like so:
```kotlin
class MyTest : TearDown() {
```
When extending this class, state is automatically cleaned up between tests. If needed, you can override the `tearDown` method but make sure to call its super-implementation.

### 2. Tests using FirestoreReferenceList
For tests calling `FirestoreReferenceList.requestAll()`, it is crucial to make a static mock of `Firebase.firestore`. Otherwise the test will try to make a real request, which will make the end to end tests fail (see aforementionned issue for more information).


app/src/androidTest/java/com/android/unio/Utils.kt
```diff
@@ -8,17 +8,22 @@ import androidx.compose.ui.test.onNodeWithTag
...
+import org.junit.After
 
 /*
  * Scrolls to a component if it's not displayed and asserts if it is displayed
  */
-
-fun assertDisplayComponentInScroll(compose: SemanticsNodeInteraction) {
-  if (compose.isNotDisplayed()) {
-    compose.performScrollTo()
+fun SemanticsNodeInteraction.assertDisplayComponentInScroll() {
```
Romainhir, line 22: Is there a reason why you made this change ?

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
```diff
@@ -13,14 +18,26 @@ import com.google.firebase.auth.auth
...
 open class EndToEndTest : FirebaseEmulatorFunctions {
+  init {
+    assertEquals(
+        """There are still listeners attached to the Auth instance. Make sure to remove 
+           them between tests with Firebase.auth.unregisterAllAuthStateListeners().
+        """
+            .trimIndent(),
+        0,
+        Firebase.auth.currentAuthStateListenerCount())
+  }
```
Zafouche, line 37: Is there no way of removing the remaining listeners in this `init` block, if they are present?

app/src/main/java/com/android/unio/model/authentication/AuthStateListeners.kt
```diff
@@ -0,0 +1,18 @@
...
+  authStateListeners.add(callback)
+  this.addAuthStateListener(callback)
+}
+
+fun FirebaseAuth.unregisterAllAuthStateListeners() {
+  authStateListeners.forEach { this.removeAuthStateListener(it) }
+  authStateListeners.clear()
+}
+
+fun FirebaseAuth.currentAuthStateListenerCount(): Int = authStateListeners.size
```
Zafouche, line 18: Quite simple yet very useful!

Zafouche (approved):
Huge fix, great work! 🚀

This will make our life much easier, and will help us lose less time on obscene bugs while testing. 😅

This makes a lot of improvements possible, like removing function calls like `assertDisplayComponentInScroll ` before checking if the node is displayed, which should be anyway, so a much needed fix.
 
Also, great documentation throughout the process!

app/src/androidTest/java/com/android/unio/Utils.kt
oskar-codes: Yes. Although the behavior is still the same, this method signature avoids passing a SemanticsNodeInteraction as parameter to another function because that is generally considered bad practice, as the method thus would share state between tests.

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
oskar-codes: Yes, that assertion is made to check that no listeners are currently attached to the auth instance. If there are any, it means that they might still hold state from previous tests, which is something we want to avoid.

However, removing auth listeners is already handled by `TearDown()`, so all you need to do is extend that class in your tests and this assertion should never fail.

Also note that the assertion is not called in a `@Before` method because at that point the new listeners from `MainActivity` have already been created so the assertion would always fail, but when asserted from `init` we ensure that we only check the listeners of the previous tests.

app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
```diff
@@ -301,9 +327,9 @@ class AssociationProfileTest {
...
   }
 
-  @After
-  fun tearDown() {
-    clearAllMocks()
-    unmockkAll()
+  @Module
+  @InstallIn(SingletonComponent::class)
+  object FirebaseTestModule {
+    @Provides fun provideFirestore(): FirebaseFirestore = mockk()
```
Redd87, line 333: What does this do?

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
```diff
@@ -78,9 +100,13 @@ open class EndToEndTest : FirebaseEmulatorFunctions {
       Firebase.firestore.useEmulator(HOST, Firestore.PORT)
       Firebase.auth.useEmulator(HOST, Auth.PORT)
       Firebase.functions.useEmulator(HOST, Functions.PORT)
-      println("Firebase emulators configured successfully.")
     } catch (e: IllegalStateException) {
-      // Firebase emulators have already been initialized. Skipping configuration.
+      Log.d("EndToEndTest", "Firebase Emulators are already in use. $e")
```
Redd87, line 104: Maybe change this Log.d to a Log.e?

Redd87 (commented):
Good work, this will make the code a lot cleaner!

app/src/androidTest/java/com/android/unio/end2end/EndToEndTest.kt
Romainhir: It should maybe be a warning or an info log

PR #210, opened 2024-11-25
feat: implement Places Search API 

Zafouche:
## ⚠️ This PR will be abandoned (or not entirely at least)
Problem: Unfortunately, I have found no way of bypassing the requirement of a billing address in the Google cloud services, so this PR will most likely be abandoned. Because of this, I unfortunately cannot see the results of my work, only that it makes the queries to the Places Search API, but the response never returns predictions, only the error message saying that a billing address is required.

## Instead 
We have to turn to a completely free (and hassle-free) solution. In the bootcamp, we had used `Nominatim`, but I think it is a poor library, at least, compared to other solutions. After some research, I have found that people prefer the solution of `mapbox`'s [Search box API](https://docs.mapbox.com/api/search/search-box/). Try playing around with their [playground example](https://docs.mapbox.com/playground/search-box/). It shows it can fetch really nice informations about locations, and that we can essentially reproduce something similar to the average map apps out there and easy to the eye.

## Description

This PR is an attempt at implementing the integration of the Google Places API into the `Event` creation screen. It allows users to search for and select a location using the Places Autocomplete API. The search bar is essentially similar to the search bar in the Googles Maps app. The location suggestions are fetched as the user types in the location outlined text field and allows them to select a place from a dropdown, filling the location with the place’s name and address.

## Motivation and Context

By integrating the Google Places API, we enable users to search and select a location using the Autocomplete feature, to make Location selection very easy and similar to what most people are used to, e.g. Google Maps. 

### PlacesSearchViewModel:
- Added logic to query the Places API for location suggestions as the user types in the search field.
- The query is sent to the repository, which returns a list of autocomplete predictions (locations).
- When a user selects a prediction, more data about it is fetched and the actual Location object is built.

### PlacesSearchRepository:
- Used the PlacesClient to query for autocomplete predictions based on the user’s search query.
- Fetches information (name, address and coordinates) of a selected place using its Place ID, and builds a `Location` object.

### EventCreationScreen:
- The location field displays a list of autocomplete suggestions as the user types.
- Integrated the PlacesSearchViewModel to handle fetching of location suggestions.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -187,6 +187,7 @@ dependencies {
     implementation(libs.core)
     implementation(libs.firebase.functions.ktx)
     implementation(libs.play.services.location)
+    implementation(libs.places)
 
     testImplementation(libs.test.core.ktx)
     debugImplementation(libs.androidx.ui.tooling)
--- a/app/src/androidTest/java/com/android/unio/ui/ScreenDisplayingTest.kt
+++ b/app/src/androidTest/java/com/android/unio/ui/ScreenDisplayingTest.kt
@@ -252,7 +252,12 @@ class ScreenDisplayingTest {
   @Test
   fun testEventCreationDisplayed() {
     composeTestRule.setContent {
-      EventCreationScreen(navigationAction, searchViewModel, associationViewModel, eventViewModel, placesSearchViewModel)
+      EventCreationScreen(
+          navigationAction,
+          searchViewModel,
+          associationViewModel,
+          eventViewModel,
+          placesSearchViewModel)
     }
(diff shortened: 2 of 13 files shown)


Zafouche:
Note that if you try to run this, you have to add the following line in the `onCreate` function of the `MainActivity`:
```kotlin
Places.initialize(context, "YOUR_API_KEY")
``` 
where `YOUR_API_KEY ` is an API key which has the Places API enabled. With a key that enables it while skipping the billing address confirmation, you will reproduce the error saying that a billing address is required when you try to make a query, e.g. type into the `Location` outlined text field some text.

Zafouche:
I will close this PR, but will keep the branch for now.

Issue #212, opened 2024-11-25
Adapt the "following" feature to work in offline mode


Romainhir:
There are multiple feature that do not work in offline mode. The "following" feature is one of them.

The user has to leave the screen and come back for the button to be clickable again, and the follow count isn't increased. 
When you follow an association, two repositories are essentially called in ConcurrentAssociationUserRepositoryFirestore, which runs the calls in a runBatch. 
I suspect this batch operation is the issue, and has to be handled differently to work with offline mode, or another operation should be done after to fix the bug.

Romainhir:
I managed to identify the problem: when executing a batch operation (multiple write to firestore) in offline mode, it does not trigger an “onComplete” event (i.e. neither “onSuccess” nor “onFailure” is executed). This means that the given callbacks are not executed. 
However, the event is triggered if writes are made to the actual database (and not to the cache), which means that when we're online, the callback functions are executed. But this can cause synchronization problems because some of the parameters given to the callback are wrong.

Romainhir:
With our current implementation of the follow system (that block the follow button when clicking on it and unblock in a callback) it seems hard to implement the follow mechanism in offline mode without messing up the synchronization of datas. Thus, I made that the button display a Toast when the user is in offline mode and nothing else, preventing the issues described above.

PR #214, opened 2024-11-25
Adapt the "following" feature to work in offline mode

Romainhir:
There are multiple feature that do not work in offline mode. The "following" feature is one of them.

The user has to leave the screen and come back for the button to be clickable again, and the follow count isn't increased.
When you follow an association, two repositories are essentially called in ConcurrentAssociationUserRepositoryFirestore, which runs the calls in a runBatch.
I suspect this batch operation is the issue, and has to be handled differently to work with offline mode, or another operation should be done after to fix the bug.

### DIFF
--- a/app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
+++ b/app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
@@ -258,27 +258,27 @@ class AssociationProfileTest : TearDown() {
     assert(associationViewModel.selectedAssociation.value!!.followersCount == currentCount)
   }
 
-    @Test
-    fun testFollowOffline() {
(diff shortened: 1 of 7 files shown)


Romainhir:
See #212 for more details

AlouchLaBouche (approved):
Nice fix ! I think it is also present in your other PR for the [password management](https://github.com/SwEnt-Group13/Unio/pull/219). 
Anyhow Looks good to me !

Romainhir:
> Nice fix ! I think it is also present in your other PR for the [password management](https://github.com/SwEnt-Group13/Unio/pull/219). Anyhow Looks good to me !

It's a bug because I implemented these PR concurrently. This should be removed from the other PR when this one will be merged on main.

PR #219, opened 2024-11-27
Implement password reset functionnality

Romainhir:
Currently, once an account has been created, it is not possible to change the password in the app. This task aims to ass password reset functionnality on the account profile of the user. Precisely, this will send an email to the user with a link where they can modify his password. This does not allow a user to change it in the Welcome Screen. This can be done in another task.

Edit : I also added the task to reset the password from the Welcome screen in this PR, because it makes more sens to me to merge them in the same time


app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
AlouchLaBouche, line 1: Great End to End test :rocket: !

app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    signInWithUser(composeTestRule, MarjolaineLemm.EMAIL, MarjolaineLemm.NEW_PASSWORD)
+
+    composeTestRule.waitUntil(10000) {
+      composeTestRule.onNodeWithTag(HomeTestTags.SCREEN).isDisplayed()
+    }
+
+    signOutWithUser(composeTestRule)
+  }
+
+  private fun simulateResetPassword() {
```
AlouchLaBouche, line 71: You might want to extract this method, this will avoid duplication, Or even simply combine the two tests files `ResetPasswordSettingsTest` and `ResetPasswordWelcomeTest` as they are technically very similar.

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
```diff
@@ -143,7 +144,32 @@ fun WelcomeScreen(userViewModel: UserViewModel) {
                       else PasswordVisualTransformation(),
               )
 
-              Spacer(modifier = Modifier.size(70.dp))
+              TextButton(
+                  onClick = {
+                    authViewModel.sendEmailResetPassword(
+                        email,
```
AlouchLaBouche, line 150: Watch out ! I tested your implementation and if email is null (the user hasn't written anything) the app will crash ! 
You should check the value of the email variable before sending any reset password, typically this can be done with the `isValidEmail` function located in `Auth.kt`!

app/src/main/java/com/android/unio/ui/settings/Settings.kt
```diff
@@ -169,6 +197,21 @@ fun SettingsContainer() {
                       Manifest.permission.ACCESS_COARSE_LOCATION))
             }
           })
+      preference(
```
AlouchLaBouche, line 200: I personally believe a toast to indicate that a link  has been sent to reset the password is a better idea than the added text, and goes better with the rest of the app. But this is personal preference and your implementation works fine.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -281,13 +283,21 @@ private fun AssociationProfileContent(
...
   }
   var enableButton by remember { mutableStateOf(true) }
+  val isConnected = Utils.checkInternetConnection(context)
+
   val onFollow = {
-    enableButton = false
-    associationViewModel.updateFollow(association!!, user!!, isFollowed) {
-      userViewModel.refreshUser()
-      enableButton = true
+    if (isConnected) {
```
AlouchLaBouche, line 289: Nice fix to check whether the user is connect or not to ensure correct following tracking

AlouchLaBouche (changes requested):
Overall great work :fire: ! This feature is very important for the user experience. However I will request changes due to the fact that when on the welcome screen it takes as value the email of the outlined text field and if it's null the app crashes. I personally believe that a separate screen should be made for the reset password, as it is done in other applications.

AlouchLaBouche:
By the way we noticed with @oskar-codes that there exists a method to change the password with firebase auth `updatePassword` see the [documentation](https://firebase.google.com/docs/auth/android/manage-users?_gl=1*3bpwg*_up*MQ..*_ga*MTg4ODI4MjMxNy4xNzMyNzMzOTE3*_ga_CW55HF8NVT*MTczMjczMzkxNy4xLjAuMTczMjczMzkxNy4wLjAuMA..#set_a_users_password). This can be a better use for changing the password when signed in.
Furthermore, I have already done the UI for the forgot password screen, see the  [figma](https://www.figma.com/design/rlvqRyIsbdSyEbXu8Xp7yG/Unio?node-id=1438-4065&node-type=frame&t=lvOu7g6G7JazMRZf-0), so we can quickly implement it and merge everything to main.

app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
Romainhir: I don't know how two e2e tests are handled in the same file but I can definitely extract the function.

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
Romainhir: Yes, you are right, my bad

Romainhir:
> By the way we noticed with @oskar-codes that there exists a method to change the password with firebase auth `updatePassword` see the [documentation](https://firebase.google.com/docs/auth/android/manage-users?_gl=1*3bpwg*_up*MQ..*_ga*MTg4ODI4MjMxNy4xNzMyNzMzOTE3*_ga_CW55HF8NVT*MTczMjczMzkxNy4xLjAuMTczMjczMzkxNy4wLjAuMA..#set_a_users_password). This can be a better use for changing the password when signed in. Furthermore, I have already done the UI for the forgot password screen, see the [figma](https://www.figma.com/design/rlvqRyIsbdSyEbXu8Xp7yG/Unio?node-id=1438-4065&node-type=frame&t=lvOu7g6G7JazMRZf-0), so we can quickly implement it and merge everything to main.

So how do you see the thing ? Should I still merge this to main and we will make the change later or should we put our code in the same PR ?

app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    val ip = "10.0.2.2"
```
oskar-codes: Consider using the already existing `HOST` constant.

app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
oskar-codes: Just so you know, having two test classes in the same file produces the same behavior as having two separate files containing each class.
Also, if you chose extracting the method, consider simply placing it in the `EndToEndTest` class so that all other end to end tests can access it.

AlouchLaBouche:
I think we can see each other tomorrow @Romainhir and quickly add my `forgot password screen` UI to your PR and then merge it :+1: .

app/src/androidTest/java/com/android/unio/end2end/ResetPasswordSettingsTest.kt
Romainhir: Thank you for the clarification

AlouchLaBouche (approved):

Zafouche (approved):
LGTM

PR #231, opened 2024-12-01
Chore: Make the app completely translated to French.

armouldr:
There still were some strings that didn't translate to English, this PR aims to translate them all. 

During this process I discovered the following bug : When we changed the language through our settings screen and then restarted the app, the app would launch in the system's language (not taking the change of settings into account). It would only display the correct language when travelling back to Unio's settings screen.
As a fix, I moved the code responsible for changing the Locale according to the language setting from Settings.kt to MainActivity.kt and that did the trick :) Still took me a bit of time to understand how this worked before finding the fix.


Zafouche (approved):
LGTM. Nothing to say, much needed refactor again! 

Also nice catch moving the language setter to the our main composable, such that it is always correctly set at the beginning and not only when we go in the settings screen.

Zafouche:
Could you merge this? It's been approved for a while now.

Issue #228, opened 2024-11-29
Continue to check for features that don't work in offline mode [CONTINUE IN SPRINT 8]


Zafouche:
## **[SEE NEXT COMMENT FOR FULL INVESTIGATION, THIS ONE IS OUTDATED]**

A couple of issues were raised in [this issue](https://github.com/SwEnt-Group13/Unio/issues/201):

Offline mode should work everywhere in the project, in theory(read and write queries). 
However I noticed that some operations do not work in offline mode:

- The following feature: (the user has to leave the screen and come back for the button to be clickable again, and the follow count isn't increased). When you follow an association, two repositories are essentially called in `ConcurrentAssociationUserRepositoryFirestore`, which runs the calls in a `runBatch`. I suspect this batch operation is the issue, and has to be handled differently to work with offline mode, or another operation should be done after to fix the bug. [**HANDLED**]
- Then, the event creation also does not work in offline mode. This might be an issue with the emulator, and should be tested on an actual phone, but the event creation isn't sent when connectivity is regained, at least when I tested on the emulator, so that needs further investigation. EDIT: I just forgot and just realized, this is normal and @Redd87 implemented it, it's about to be merged!
- During event creation, the dialogs that display association to tag or associate to the event does not work w/o connectivity. This should work, because the list of associations is loaded in the respective view model anyway.
- Other operations remain to be tested further, I doubt I have covered every case.

Zafouche:
I'm currently testing this from scratch, I will collect in this comment what I find: 

- Cannot create an event while offline: this is to be discussed, if we want a user to be able to create an event offline. If we do, that means we should allow the user to give, to the location input, whatever he wants to write in. This means the user can also choose to not write an address, and put in whatever word or sentence they want. In offline mode, it wouldn't be possible to restrict this input field to a `Location`, which is pretty central to an `Event`, or else how would we know where the event is. Either we allow the user to create an event w/o a location, or just disallow the user to create an event w/o connection. **[TO DISCUSS]**

Zafouche:
I will keep this issue open so that we can give updates directly to this issue when we fix something.

Zafouche:
I've read that, in offline mode, the callbacks are executed for read operation, but never for write operations as they wait for the server's response to complete.

I've also been looking at how others find a fix for this issue (very few). 

All similar issues with potential solutions I see lead to what's described [here](https://github.com/firebase/firebase-android-sdk/issues/642). Essentially, they either : 
[...]

PR #234, opened 2024-12-02
Fix/edit profile offline

Romainhir:
- Description modification: This works, but with a bug. When we change the description and press on the save button, nothing seems to happen, e.g. the app doesn't take the user back to the MyProfile screen automatically. However, the description is successfully modified, and I can see my new description when switching screen manually by clicking on the Back button. So it works, but this behavior should be fixed.
- First name and last name modification: works, but with the same buggy behavior as above.
- Social media: works, same as above
- Interests: same as above.

When we gain back connection, it indeed uploads to changes made in offline mode to firebase.

### DIFF
--- a/app/src/androidTest/java/com/android/unio/components/user/UserProfileEditionTest.kt
+++ b/app/src/androidTest/java/com/android/unio/components/user/UserProfileEditionTest.kt
@@ -1,5 +1,7 @@
 package com.android.unio.components.user
 
+import android.net.ConnectivityManager
+import android.net.Network
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.junit4.createComposeRule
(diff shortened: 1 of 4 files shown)


Romainhir:
Note that I cannot have a better coverage because I cannot disable network in a end-to-end test. And in a UI test, mocking data like in "UserProfileEditionTest" seems to not increase the coverage although the case is tested.

Zafouche:
For offline mode, I've seen people mention countless times using `SnapshotListeners` countless times, have you already looked at that? I still need to do more research on the topic so I'm not sure yet if we could use that, but it seems like it is something worth exploring.

Romainhir:
> For offline mode, I've seen people mention countless times using `SnapshotListeners`, have you already looked at that? I still need to do more research on the topic as I'm not sure yet if we could use that, but it seems like it is something worth exploring.

Thank you for the suggestion. I read more about "SnapshotListeners". This could be helpful for issue #235 .Anyway, I'd be glad to know what you think about the fix proposed in this RP.

Zafouche:
All other issues with potential solutions I see lead to what's described [here](https://github.com/firebase/firebase-android-sdk/issues/642). Essentially, they either : 
- Execute their `onSuccess` callbacks after having executed `ref.set(value)`, because the write is executed on the local cache whether you are online or offline. We should have some thought about this to figure out how we want to separate offline callbacks and online callbacks.
- Use `addValueEventListener` on the `.info/connected` ref, documented [here](https://firebase.google.com/docs/database/android/offline-capabilities#section-connection-state). We should have a look at that but I don't know if it would be applicable in our case yet.

Zafouche:
If someone else in the team could give their opinions on this, feel free to do so. It would be helpful for us, so we can decide how we want to proceed for all issues concerning offline mode described in [this issue](https://github.com/SwEnt-Group13/Unio/issues/228).

app/src/androidTest/java/com/android/unio/components/user/UserProfileEditionTest.kt
```diff
@@ -40,16 +56,116 @@ class UserProfileEditionTest : TearDown() {
     navigationAction = mock(NavigationAction::class.java)
```
Redd87, line 56: Try not to mix Mockk and Mockito

app/src/androidTest/java/com/android/unio/components/user/UserProfileEditionTest.kt
```diff
@@ -142,4 +270,10 @@ class UserProfileEditionTest : TearDown() {
         .onNodeWithTag(UserEditionTestTags.LAST_NAME_ERROR_TEXT, useUnmergedTree = true)
         .assertExists()
   }
+
+  object UserUpdate {
```
Redd87, line 274: 😎

Redd87 (approved):
Not much to say, some minor comment about mockk and mockito but mockito doesn't come from your code, still consider refactoring it if you have time. Otherwise good job LGTM

Issue #235, opened 2024-12-02
Make search bars work (at normal speed) in offline mode


Romainhir:
For full offline mode feature overview, see [this issue](https://github.com/SwEnt-Group13/Unio/issues/228).

## Search bars:
- `Home` Screen: a search query takes anywhere from 30-60 seconds personally. It looks like there might be caching issues, or the data is not properly stored, but whatever it is, it makes the search queries **muchhhh** slower.
- `Explore` Screen: similar to the `Home` Screen. From my testing, it took a minute for the first query, then 5-10 seconds for the second query, then 20 for the third,... Basically, it's also way too long. Seems there might be an issue with where or how the search view model gets its data. 

### Possible solutions: 
Check the implementation of the whole search bar mechanism.

Romainhir:
This is quite similar to #212 and #233. The problem also comes from the callback of the Firestore operation. But here, the callback is executed correctly (the operation is marked as “Success”) but has to wait a while before being executed. This is because Firestore tries to reach the database and never receives a response. This causes a timeout, at which point the data is read from the cache.
This bug is trickier to eliminate because the callbacks are correctly executed, unlike #212 and #233. We have to read the database for the search to work. 
The good news is that the search behaves correctly. It's “just” very slow.

Romainhir:
![image](https://github.com/user-attachments/assets/bc5cd1b6-67ea-4547-94d8-65c29a98aade)

A small image to illustrate what I said above. The logs are executed just before the “get” operation for the database, and at the start of the “Success” callback.

Romainhir:
Query duration is proportional to the number of results they contain. Requests with few results (1 or 2) take a normal amount of time, but longer ones (with 10 results, for example) take much longer. An example of a long query is “age” or “agepoly”: there are 10 results and you have to wait a few minutes for the results to appear.

PR #236, opened 2024-12-02
Make search bars work (at normal speed) in offline mode

Romainhir:
For full offline mode feature overview, see https://github.com/SwEnt-Group13/Unio/issues/228.

### Search bars:

- Home Screen: a search query takes anywhere from 30-60 seconds personally. It looks like there might be caching issues, or the data is not properly stored, but whatever it is, it makes the search queries muchhhh slower.
- Explore Screen: similar to the Home Screen. From my testing, it took a minute for the first query, then 5-10 seconds for the second query, then 20 for the third,... Basically, it's also way too long. Seems there might be an issue with where or how the search view model gets its data.

### Possible solutions:

Check the implementation of the whole search bar mechanism.


app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
```diff
@@ -72,11 +74,20 @@ class AssociationRepositoryFirestore @Inject constructor(private val db: Firebas
...
+        onFailure(exception)
+        return@registerSnapshotListener
+      }
+      if (documentSnapshot != null && documentSnapshot.exists()) {
+        onSuccess(hydrate(documentSnapshot.data))
+      }
+    }
+    //        .performFirestoreOperation(
+    //            onSuccess = { document -> onSuccess(hydrate(document.data)) },
+    //            onFailure = { exception -> onFailure(exception) })
```
Zafouche, line 90: Commented code should be removed.

app/src/main/java/com/android/unio/model/firestore/FirestoreUtils.kt
Zafouche, line 1: Nicely done.

app/src/main/java/com/android/unio/ui/user/UserClaimAssociationPresidentialRights.kt
```diff
@@ -181,8 +180,6 @@ fun UserClaimAssociationPresidentialRightsScreenScaffold(
                                                   .user_claim_association_presidential_rights_unexpected_error),
                                           e)
                                     }
-                                  } else {
-                                    // it works !
```
Zafouche, line 185: Interesting, I never noticed this comment, good catch.

app/src/test/java/com/android/unio/model/association/AssociationRepositoryFirestoreTest.kt
Zafouche, line 1: Good job on making this test only use one of `MockK` or `Mockito` (`MockK` in this case).

Zafouche (commented):
Clean fix, nothing to say, just remove the commented code and let me know when that's done. 

Since only both of us worked on offline mode I'll just leave this as a comment, and it seems someone else will review this later, so you should wait for that to merge this. 

Also just a curiosity question, what do the changes in the firestorm emulator files mean? Did you simply update the emulator? (it prompts me with an update suggestion every time I run it that's why I'm asking)

Oh and nice test addition for the offline mode! I see you ended up finding the solution to disable internet in the test specifically.

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
Romainhir: You are right, my bad

app/src/main/java/com/android/unio/model/firestore/FirestoreUtils.kt
```diff
@@ -2,7 +2,26 @@ package com.android.unio.model.firestore
...
 import com.google.android.gms.tasks.Task
+import com.google.firebase.firestore.DocumentReference
 import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.EventListener
+import com.google.firebase.firestore.ListenerRegistration
+import com.google.firebase.firestore.MetadataChanges
+
+private val snapshotListeners = mutableListOf<ListenerRegistration>()
+
+fun DocumentReference.registerSnapshotListener(
```
Redd87, line 13: Don't forget to comment your functions!

app/src/main/java/com/android/unio/model/firestore/FirestoreUtils.kt
```diff
@@ -2,7 +2,26 @@ package com.android.unio.model.firestore
...
+    metadata: MetadataChanges,
+    callback: EventListener<DocumentSnapshot>
+) {
+  val listener = this.addSnapshotListener(metadata, callback)
+  snapshotListeners.add(listener)
+}
+
+fun unregisterAllSnapshotListeners() {
```
Redd87, line 21: Here as well

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
```diff
@@ -72,11 +74,17 @@ class AssociationRepositoryFirestore @Inject constructor(private val db: Firebas
...
+        documentSnapshot,
+        exception ->
+      if (exception != null) {
+        onFailure(exception)
+        return@registerSnapshotListener
+      }
+      if (documentSnapshot != null && documentSnapshot.exists()) {
+        onSuccess(hydrate(documentSnapshot.data))
+      }
+    }
```
Redd87, line 87: As I didn't work on the offline mode, I'm not sure I understand what this does or what the fix is? Could you maybe comment this some more?

Redd87 (changes requested):
Good job on fixing this ! Just add a few comments here and there for future maintainability. And good job as well on migrating the tests from Mockito to MockK

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
Romainhir: Check #235 and #228 for a description of the issue. Here, instead of using success and failure listener directly, we use a Snapshot Listener that call directly the callback when a read/write are made on the local (cache) database. This prevent this feature to be stuck for a long time without result. See [this](https://firebase.google.com/docs/firestore/query-data/listen?hl=fr) for more info

Redd87: Ok thanks!

Redd87 (approved):
Good job!

PR #240, opened 2024-12-03
Feature/edit with nominatim

Redd87:
As my previous PR adding the event edit feature was developed concurrently with @Zafouche 's PR that added the nominatim location picker in the EventCreation composable, the location picker wasn't added to the Edit screen. This PR aims to fix this. It extracts the location picker in the EventEditComponents file and changes EventCreation.kt accordingly, and adds it to EventEdit. I also added some documentation to EventEditComponents as it was lacking. 

This is currently a draft PR as further tests are needed in EventEdit.

### DIFF
--- a/app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
+++ b/app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
@@ -60,6 +60,9 @@ import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 
+const val DROP_DOWN_MAX_CHARACTERS = 40
+const val DROP_DOWN_MAX_ROWS = 3
+
 @OptIn(ExperimentalLayoutApi::class)
 @Composable
 fun AssociationChips(
--- a/app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
+++ b/app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
@@ -78,6 +78,7 @@ const val DROP_DOWN_MAX_ROWS = 3
  *
  * @param locationSearchViewModel NominatimLocationSearchViewModel : ViewModel for searching
  *   locations.
+ * @param initialLocation Location? : Initial location to pre-fill the text field.
  * @param textFieldTestTag String : Test tag for the text field.
  * @param dropdownTestTag String : Test tag for the dropdown menu.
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
```diff
@@ -50,16 +56,103 @@ import androidx.compose.ui.platform.LocalContext
...
+                  showDropdown = false
+                },
+                modifier = Modifier.padding(8.dp).testTag(dropdownTestTag + location.latitude))
+            HorizontalDivider()
+          }
+
+          if (locationSuggestions.size > DROP_DOWN_MAX_ROWS) {
+            DropdownMenuItem(
+                text = { Text(context.getString(R.string.event_creation_location_dropdown_more)) },
+                onClick = {},
```
Romainhir, line 143: Empty onClick parameter. Is there really nothing to do on click here (not even a Toast or something) ?

Romainhir (changes requested):
Nice job ! I only have one question otherwise, LGTM.

app/src/main/java/com/android/unio/ui/components/EventEditComponents.kt
Zafouche: Yeah this comes from the implementation that was done in my PR of the `Location` input in the `Event` creation screen. which he simply refactored but did not modify it in anyway (or at least I don't see anything that has been modified).

@Redd87 I suggest you add a `Toast` telling the user to continue writing the address, such that the one that he wants shows up in the first three suggestions. 

Or you can make it such that it adds more suggestions to the `Dropdown`, but I don't think allowing the user to extend the Dropdown infinitely is a good idea. 

We could refine it later on by allowing to add a couple of suggestions, only once, but this is not important for this PR.

The first solution, IMO, is the best, which simply consists in adding a `Toast`. 
Please add it if possible, otherwise LGTM as well.

Romainhir: Ok, I'll approve for now then. Thank you for the answer

Zafouche: Still it'd be great if @Redd87 could add the `Toast` before merging.

Redd87: @Zafouche ok I'll let you add the toast

PR #243, opened 2024-12-04
Make saving an event faster and smoother (both online and offline mode)

Romainhir:
While testing offline mode, I noticed how slow this mechanism was. But also, the normal backend, even when not in offline mode, seems to work slowly. I think the heart should instantly turn red, and then saved quickly such that you instantly see it appear in the Saved screen. But it was especially slow in offline mode.

For full offline mode feature overview, see https://github.com/SwEnt-Group13/Unio/issues/228.
### Saving an event:

When we press the heart button to save an event, it takes 3-4 seconds for it to be processed and show up in the Saved screen.
### Possible solutions:

Not sure. The backend of the saving mechanism needs to be investigated to determine what causes it to be slow. But it works! But its slow.


app/src/main/java/com/android/unio/ui/event/EventCard.kt
```diff
@@ -167,6 +165,10 @@ fun EventCard(
       shouldBeEditable = shouldBeEditable)
 }
 
+private operator fun Any.getValue(nothing: Nothing?, property: KProperty<*>): Any {
+  TODO("Not yet implemented")
+}
```
Zafouche, line 170: Unused function If I'm not mistaken? and no TODOs in the code please

app/src/test/java/com/android/unio/model/user/UserRepositoryFirestoreTest.kt
```diff
@@ -65,14 +65,23 @@ class UserRepositoryFirestoreTest {
 
   @Before
   fun setUp() {
-    MockitoAnnotations.openMocks(this)
+    //    MockitoAnnotations.openMocks(this)
```
Zafouche, line 68: Commented code.

app/src/test/java/com/android/unio/model/user/UserRepositoryFirestoreTest.kt
Zafouche, line 1: Yet again, good job on fully shifting this to `MockK` instead of mixing it with `Mockito`.

Zafouche (changes requested):
Small but effective fix.
Great job on this, it overall LGTM, I just need you to make two small changes and I'll gladly approve that asap. 

Solid line coverage, just under the 80%, but since this PR is pretty much just refactoring, this is not an issue. And great job on adapting the tests to the shift to these snapshot listeners!

app/src/main/java/com/android/unio/ui/event/EventCard.kt
Romainhir: Oh I did not see this. I think Android Studio automatically added this. Thank you !

app/src/test/java/com/android/unio/model/user/UserRepositoryFirestoreTest.kt
Romainhir: Forgot to remove it, my bad

Zafouche (approved):
Perfect thanks for making the changes, LGTM.

PR #245, opened 2024-12-04
Feature/number of saved

Redd87:
One of the key features of the app is to improve the organization of an event from the association pov, which included seeing how many people plan on going to an event.  This PR aims to introduce this in the application.  As the initial design did not specify how to implement this, I simplified it to Event Saved = Person is going after discussion with the team, thus this PR includes:
- a new property for the Event type NumberOfSaved
- updated serialization and hydration methods
- modification to the onClickSaved in eventCard to modify the event and account for the additional saved

This PR is currently a draft as these features have not yet been extensively tested


Redd87:
While I was there, I took the occasion to fix the search bar expanded state; it now properly closes when the user clicks outside of it

Redd87:
Two further bugs were fixed: the hydration of an event improperly hydrated int values which resulted in default values being used and the updateEvent of the EventViewModel resulted in the ID of the event being duplicated in the Events parameter of the association. I also found another bug which I did not fix but if an association has more than 30 events it will not load them on the association profile

Redd87:
The PR is open for review. The lack of coverage for the low number of lines is due to the EventDetails not being tested, but as it needs to be rewritten to account for the notifications handler which haven't been added there yet I consider that it should be done in another PR

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
Zafouche, line 1: It's good that you added quick checks for the new property in the hydration and serialization mechanisms 👍

app/src/main/java/com/android/unio/ui/event/EventDetails.kt
```diff
@@ -110,11 +110,22 @@ fun EventScreen(
...
+          newEvent,
+          onSuccess = {},
+          onFailure = { e -> Log.e("EventCard", "Failed to update event: $e") })
     } else {
       user!!.savedEvents.add(event!!.uid)
+      val newEvent = event!!.copy(numberOfSaved = event!!.numberOfSaved + 1)
+      eventViewModel.updateEventWithoutImage(
+          newEvent,
+          onSuccess = {},
+          onFailure = { e -> Log.e("EventCard", "Failed to update event: $e") })
```
Zafouche, line 128: Testing this mechanism in the UI test shouldn't be very difficult and would put your line coverage above 80% for sure. Do you think you could do that? Since it's pretty much the feature that you implemented, I do think it would deserve a test.

Zafouche (commented):
I haven't caught any absurdities, the code looks good to me. 👍
Simple and efficient, well done.

As for the tests, adding a test that checks the feature mechanism in `EventDetails` would be good I think, check my comment below. It should easily raise the line coverage above the 80% mark.

app/src/main/java/com/android/unio/ui/event/EventDetails.kt
Redd87: I agree that it isn't very difficult but it feels absurd to add a whole UI test for something that should be refactored ASAP, but if you don't agree I'll get on it right now

Redd87: That would have been the plan initially but we have to see if we're going to have time to develop that with the time remaining. But what I meant by this needs to be refactored is specifically the onClickSaveButton function which is not complete here in EventDetails: it doesn't use the appropriate UserViewModel methods and does not register the notification handler, both of which are out of the scope of this PR and will require the function to be rewritten

Zafouche: Ah yes you had mentioned it in your last comment, that's fine then 👍

Zafouche (approved):
LGTM.

Issue #248, opened 2024-12-05
Disallow event creation in offline mode.


Zafouche:
For full offline mode feature overview, see [this issue](https://github.com/SwEnt-Group13/Unio/issues/228).

For this specific feature, feel free to discuss what solution you would feel is good under this issue.

## Event creation: 
We should decide if we want to allow a user to create an event while offline. If we do, that means we should allow the user to give, to the location input, whatever he wants to write in. This means the user can also choose not to write an address, and put in whatever word or sentence they want. In offline mode, it wouldn't be possible to restrict this input field to a `Location`, which is pretty central to an `Event`, or else how would we know where the event is. 

### Possible solutions:
 Either we allow the user to create an event w/o a location, just disallow the user to create an event w/o connection, or trust the user to put an actual address in the location field. **[USER WILL BE DISALLOWED FROM CREATING AN EVENT WHILE OFFLINE]**

### **Chosen solution:**
User will be disallowed from creating an event while offline.

PR #250, opened 2024-12-05
Feat/event creation offline

Zafouche:
For full offline mode feature overview, see [this issue](https://github.com/SwEnt-Group13/Unio/issues/228).

For this specific feature, see the full description and the choice of solution behind it in [this issue](https://github.com/SwEnt-Group13/Unio/issues/248).

## Description
This PR implements a simple handling of the add event button in offline mode. When we click on it, it prevents the user from navigating to the event creation screen, which means the user isn't allowed to create one, just like a user on any social media wouldn't be allowed to create a post while offline.

## Motivation and context
This fix goes in the spirit of adapting our app's features for offline mode. This was one of the features that needed additional measures to respect it.

### AssociationProfile: 
- Added isOnline boolean that checks if the user is online or not.
- Added logic to the `Add event` button to display a toast that says the user is offline if it is the case.

## How has this been tested?

In AssociationProfileTest, I added two tests: 
- testAddEventButtonOnline: checks that the button click takes the user to the event creation screen.
- testAddEventButtonOffline: checks that the button click does *not* take the user to the event creation screen.

### DIFF
--- a/app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
+++ b/app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
@@ -9,9 +9,9 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performScrollTo
 import androidx.core.content.ContextCompat
 import androidx.core.content.ContextCompat.getSystemService
-import androidx.navigation.NavHostController
 import androidx.test.core.app.ApplicationProvider
 import com.android.unio.R
 import com.android.unio.TearDown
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/com/android/unio/components/association/AssociationProfileTest.kt
```diff
@@ -338,11 +341,9 @@ class AssociationProfileTest : TearDown() {
           navigationAction, userViewModel, eventViewModel, associationViewModel) {}
     }
 
-    `when`(navigationAction.navController.popBackStack()).thenReturn(true)
-
     composeTestRule.onNodeWithTag(AssociationProfileTestTags.GO_BACK_BUTTON).performClick()
 
-    verify(navigationAction.navController).popBackStack()
+    verify { navigationAction.goBack() }
```
AlouchLaBouche, line 346: Nice fix :+1:

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -496,7 +497,15 @@ private fun AssociationEvents(
   }
   if (isAdmin) {
     Button(
-        onClick = { navigationAction.navigateTo(Screen.EVENT_CREATION) },
+        onClick = {
```
AlouchLaBouche, line 500: Nice modification to ensure that the user is connected for before adding events !

AlouchLaBouche (approved):
Looks good to me! Nice changes that address the offline mode! Great work :rocket:

Zafouche:
> Looks good to me! Nice changes that address the offline mode! Great work 🚀

Thanks for being quick to review!

PR #257, opened 2024-12-05
Feat/add associations

Aurelien9Code:
This PR aims to create tools to fullfill the database with good-looking and existing associations, meeting a requirement said by the assistants last meeting. 

It introduces a set of utility functions in the CreateAssociationTools class for processing association data from a CSV file and saving the corresponding associations with their images. 

I'll try to make its use clear : 
- parseLineToAssociation(line: String): Association?: This function parses a CSV line containing association data (such as UID, name, category, description, and image) and converts it into an Association object.

- processAndSaveAssociations(inputStream: InputStream, saveAssociation: (Association, InputStream?) -> Unit, context: Context): This function processes a stream of CSV lines, converting each line into an Association object and attempts to save it along with its associated image (if available) using the provided saveAssociation callback.

- CSV file: The expected CSV format is as follows:
        uid,url,name,fullName,category,description,followersCount,image,principalEmailAddress
        The method skips incomplete lines and handles image resources from the raw folder.

The class is designed to be called within the MainActivity or another part of the application that initializes the tool. A sample usage snippet has been provided in the comments, demonstrating how to call the processAndSaveAssociations method asynchronously.


app/src/main/res/raw/associations.csv
Aurelien9Code, line 1: Example of csv lines, remove them before calling the program otherwise it will overwrites any changes made on these 4 associations.

misc/CreateAssociationTools.kt
```diff
@@ -0,0 +1,109 @@
...
+          association,
+          imageStream = imageStream,
+          onSuccess = { Log.i("UnioApp", "Saved association: ${association.name}") },
+          onFailure = { exception ->
+            Log.e("UnioApp", "Failed to save association", exception)
+          }
+        )
+      }, context)
+    }
+  }*/
```
oskar-codes, line 108: Consider not relying on manually adding code into MainActivity. Instead, you could use a Kotlin `main` function that doesn't even rely on running on an emulated device. 

Also, how can this code in MainActivity even work ? By default if users are not authenticated, they cannot add data to the Firestore database (is that why there is a `delay(30000)` ?). 

You should really consider making this a simple Kotlin `main` entry point, and use Firebase service accounts to bypass security rules. That would additionally greatly speed up the execution time.

oskar-codes (approved):
I will approve this pull request as it is not harmful to the project, but I believe there are better ways to do what this pull request does. See my second comment.

misc/CreateAssociationTools.kt
Aurelien9Code: Okay I can see your point and I agree; but is it really useful for this kind of utilization ? 
I will merge it so that everyone can use it to fullfill the database, but will refactor it soon with what you say.

PR #261, opened 2024-12-06
Feature [Sanitization] Add length checks to input text fields

AlouchLaBouche:
This PR adds sanitization to the input fields. Noticeably I added a new `TextLength` enum that has 3 categories : 
`LARGE`, `MEDIUM` and `SMALL` that have all associated to it a maximum character length. This ensures that when a user writes in a text field, he does not input texts that far too large for the given property and as such doesn't flood the database wiht unnecessary data. I've also added UI elements that shows up when a user is getting near the limit of the maximum text length.


oskar-codes:
Just an idea, but you could add these length checks server-side as well, in the Firestore Rules. It is fairly easy to do, I can show you at some time how.

app/src/main/java/com/android/unio/model/utils/Utils.kt
```diff
@@ -4,10 +4,32 @@ import android.content.Context
...
     return connectivityManager?.activeNetwork != null
   }
+
+  fun checkInputLength(input: String, textLength: TextLength): Boolean {
+    return when (textLength) {
+      TextLength.SMALL -> input.length <= textLength.length
+      TextLength.MEDIUM -> input.length <= textLength.length
+      TextLength.LARGE -> input.length <= textLength.length
+    }
+  }
```
oskar-codes, line 26: This is just a tip for you, but it's always useful to use Kotlin patterns that make the code more readable. For example you could rewrite this function as an extension method of `String`, which could then be used like so:
```kotlin
"Some string".hasSize(SMALL)
```
Which is somewhat more readable than the current
```kotlin
checkInputLength("Some string", SMALL)
```
This is of course subjective and up to you to include or not.

oskar-codes (approved):
Great work on this! As mentionned in my previous comment, it would be great to reflect these input limits on the server side as well. Maybe it is best to do that in another pull request, that's up to you.

app/src/main/java/com/android/unio/model/utils/Utils.kt
AlouchLaBouche: Wow okay I didn't even know about this, I'll look into it just by curiosity, thanks !

AlouchLaBouche:
I'll try and implement the server side security rules as it's literally six lines of code :+1:

PR #262, opened 2024-12-06
Feat/add role badges

Aurelien9Code:
This PR adds little colored badges next to each association member's in `UserProfile` & `AssociationProfile`. 

It also shows how roles and permissions can be used for all the project ! : ) Here, a new permission ADD_EVENT is created and the Add Event button is only showed if the permission is granted. 

As all of this is only for UI checks, there is no updates in the firestore rules. 
Next PR will be about refactoring all this !

I put some colors into constant file so that the administrator of an association can choose them and so that they are displayed equally in light or dark mode.


app/src/androidTest/java/com/android/unio/end2end/EventCreationE2ETest.kt
```diff
@@ -198,7 +198,9 @@ class EventCreationE2ETest : EndToEndTest() {
     composeTestRule.waitUntil(10000) {
       composeTestRule.onNodeWithTag(AssociationProfileTestTags.SCREEN).isDisplayed()
     }
-
+    composeTestRule.waitUntil(10000) {
+      composeTestRule.onNodeWithTag(AssociationProfileTestTags.ADD_EVENT_BUTTON).isDisplayed()
+    }
```
armouldr, line 203: Useful check :rocket:

app/src/main/java/com/android/unio/model/association/Association.kt
```diff
@@ -104,9 +109,10 @@ class Role(val displayName: String, val permissions: Permissions, override val u
                 .addPermission(PermissionType.EDIT_MEMBERS)
                 .addPermission(PermissionType.VIEW_EVENTS)
                 .build(),
+            0xFF0000FF,
             "Committee")
-    val MEMBER = Role("Member", Permissions.NONE, "Member")
-    val GUEST = Role("Guest", Permissions.NONE, "Guest")
+    val MEMBER = Role("Member", Permissions.NONE, badgeColorBlue, "Member")
+    val GUEST = Role("Guest", Permissions.NONE, badgeColorBlue, "Guest")
```
armouldr, line 115: Why did you assign the same color to all of the predefined roles ? I think it would be more useful to assign a different color to each of them.

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -393,6 +396,20 @@ private fun AssociationMembers(
...
+                    Box(
+                        modifier =
+                            Modifier.background(
+                                    color = Color(member.role.color),
+                                    shape = RoundedCornerShape(8.dp))
+                                .padding(horizontal = 8.dp, vertical = 4.dp)) {
+                          Text(
+                              text = member.role.displayName,
+                              style = MaterialTheme.typography.bodySmall,
+                              color = Color.White,
```
armouldr, line 410: You should make sure that the text color changes relatively to the color of the badge (For example, we couldn't see the text if badgeColorWhite was set for a role)

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -393,6 +396,20 @@ private fun AssociationMembers(
                   user.value?.lastName?.let {
                     val lastName = it
                     Text("${firstName} ${lastName}")
```
armouldr, line 398: A small detail, but the curly braces are not needed here

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
```diff
@@ -422,11 +439,16 @@ private fun AssociationEvents(
   var isSeeMoreClicked by remember { mutableStateOf(false) }
 
   val events by association.events.list.collectAsState()
+  val user by userViewModel.user.collectAsState()
```
armouldr, line 442: As user shouldn't be null, maybe it would be better to check if user is null and return if it is. It is what we did for other screens like EventDetails, and it allows us to write `user` instead of `user?` in the rest of the screen.

app/src/main/java/com/android/unio/ui/association/AssociationSmall.kt
```diff
@@ -23,13 +30,28 @@ import com.android.unio.ui.theme.AppTypography
...
+          if (userRole != null) {
+            Box(
+                modifier =
+                    Modifier.padding(start = 8.dp)
+                        .background(color = Color(userRole.color), shape = RoundedCornerShape(8.dp))
+                        .padding(horizontal = 8.dp, vertical = 4.dp)) {
+                  Text(
+                      text = userRole.displayName,
+                      style = AppTypography.bodySmall,
+                      color = Color.White,
```
armouldr, line 51: As the user role is present in multiple composable, I think taht it would be beneficial to move it as a separate composable. This would avoid code duplication.

app/src/main/java/com/android/unio/ui/theme/Color.kt
```diff
@@ -230,3 +230,13 @@ val eventTypeOther = Color.Gray // Default Color
...
+
+// Colors for the role badges
+val badgeColorBlack = 0xFF000000 // Opaque Black
+val badgeColorWhite = 0xFFFFFFFF // Opaque White
+val badgeColorRed = 0xFFFF0000 // Opaque Red
+val badgeColorGreen = 0xFF00FF00 // Opaque Green
+val badgeColorBlue = 0xFF0000FF // Opaque Blue
+val badgeColorYellow = 0xFFFFFF00 // Opaque Yellow
+val badgeColorCyan = 0xFF00FFFF // Opaque Cyan
+val badgeColorMagenta = 0xFFFF00FF // Opaque Magenta
```
armouldr, line 242: Maybe we will change a bit the color values in the future as full colors might be a bit painful for the eyes

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
```diff
@@ -114,18 +116,8 @@ class HydrationAndSerializationTest {
...
-        association.roles.associate { role ->
-          role.uid to
-              mapOf(
-                  Role::displayName.name to role.displayName,
-                  Role::permissions.name to
-                      role.permissions.getGrantedPermissions().map { it.stringName })
-        },
-        serialized["roles"])
+    assertEquals(mapUsersToRoles(association.members), serialized["members"])
+    assertEquals(mapRolesToPermission(association.roles), serialized["roles"])
```
armouldr, line 120: It's great to put the mapping in a separate function, makes the test class clearer :)

app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
```diff
@@ -51,11 +51,12 @@ private fun mapUsersToRoles(members: List<Member>): Map<String, String> {
...
-private fun mapRolesToPermission(roles: List<Role>): Map<String, Map<String, Any>> {
+fun mapRolesToPermission(roles: List<Role>): Map<String, Map<String, Any>> {
   return roles.associate { role ->
     role.uid to
         mapOf(
             Role::displayName.name to role.displayName,
+            Role::color.name to role.color,
             Role::permissions.name to
                 role.permissions.getGrantedPermissions().map { it.stringName })
   }
```
armouldr, line 62: This should be moved in a Utils file, but you already told me that you were planning on doing it

armouldr (changes requested):
A nice PR that adds a bit of color into our app ! There's a few little changes to be made but nothing big :rocket:

app/src/main/java/com/android/unio/model/association/Association.kt
Aurelien9Code: And it would improve my line coverage ; ) Sure

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
Aurelien9Code: Wow didn't think about that

Aurelien9Code: done !

app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
Aurelien9Code: true ! I think it was because before it was a user.firstName ; )

app/src/main/java/com/android/unio/ui/theme/Color.kt
Aurelien9Code: Yes I agree, but maybe the owner could choose it I thought

app/src/main/java/com/android/unio/ui/association/AssociationSmall.kt
Aurelien9Code: Good !

Aurelien9Code:
> A nice PR that adds a bit of color into our app ! There's a few little changes to be made but nothing big 🚀

All done !

armouldr:
Thank you for the changes it seems good to me now ! :rocket:

armouldr (changes requested):
Thank you for the changes it seems good to me now ! 🚀

armouldr (approved):

Issue #269, opened 2024-12-08
General polish and bug fixing


Redd87:
Use the application and note all bugs
Clean the database in terms of events and associations

Redd87:
Remarks so far: 
- The attributes prices, remaining places and types of an event are currently unused (always set to the default value), we should either remove them or add the ability to interact with them in the Event creation/edit
- The "I'm going" button is currently unused

Zafouche:
- [ ] For the `I'm going` button, either: 
- Remove it, and make the address button perform a click on the event marker.
- Change it such that it's a button that sends the user on the Google Maps app (like on our app, when we click on the marker, it pops up two buttons at the bottom of the map, that serve as links to the default map apps of the phone, e.g. Google Maps for android)

Zafouche:
I will now work on the following issue: 

> - The "null" in the nominatim results should be removed

This will require to define a much more complete address formatter.

Zafouche:
I will fix most of the UI inconsistencies in #280, make sure to check it before starting something UI related.

Redd87:
As the remaining problems are minor and should not bother a user during its average usage of the apps, I'm putting this in Done, but keeping the issue open for future reference

PR #270, opened 2024-12-08
docs: add missing comments to the model folder

Redd87:
This PR is linked to the issue #269, and partially fixes the lack of comments in the project. I decided to split the PR in 2: this one being the first one and focusing on the model folder and the second one focusing on the UI folder. While I was at it I removed some redundant or useless code, and also highlighted further opportunities for optimization which I linked wrote in #269.

### DIFF
--- a/app/src/main/java/com/android/unio/model/event/Event.kt
+++ b/app/src/main/java/com/android/unio/model/event/Event.kt
@@ -59,6 +59,12 @@ data class Event(
   companion object
 }
 
+/**
+ * Enum class that represents the different types of events
+ *
+ * @property color event type color
+ * @property text event type text
+ */
 enum class EventType(val color: Color, val text: String) {
   FESTIVAL(eventTypeFestival, "festival"),
   APERITIF(eventTypeAperitif, "aperitif"),
@@ -70,6 +76,17 @@ enum class EventType(val color: Color, val text: String) {
   OTHER(eventTypeOther, "other")
 }
 
+/**
+ * A class representing the event document for AppSearch indexing. It allows the search engine to
+ * search on the event title, description, catchy description and location name.
+ *
+ * @property namespace namespace of the event document
+ * @property uid event id
+ * @property title event title
+ * @property description event description
+ * @property catchyDescription event catchy description
+ * @property locationName event location name
+ */
 @Document
 data class EventDocument(
     @Namespace val namespace: String = "",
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/android/unio/model/event/Event.kt
```diff
@@ -53,12 +53,18 @@ data class Event(
     val endDate: Timestamp = Timestamp(Date()),
     val location: Location = Location(),
     val types: List<EventType> = mutableListOf(EventType.OTHER),
-    val placesRemaining: Int = -1,
+    val maxNumberOfPlaces: Int = -1,
```
Zafouche, line 56: What's this name change for? Anyways I think we don't even use this as intended. But I would agree, if that's what you thought when making this change, that it would be better to simply store the max number of places, because we can easily compute the number of places remaining from it. This prevents us from making update queries for this field every time a user signs up to the event.

app/src/main/java/com/android/unio/model/event/EventListTools.kt
Zafouche, line 1: Good call on cleaning this up, also almost unused file (except for one function).

app/src/main/java/com/android/unio/model/firestore/transform/Hydration.kt
```diff
@@ -29,7 +37,7 @@ fun AssociationRepositoryFirestore.Companion.hydrate(data: Map<String, Any>?): A
   val permissions = Permissions.NONE
   val roles =
       rolesMap.map { (roleUid, roleData) ->
-        Role.createRole(
+        Role(
```
Zafouche, line 40: Good catch 👍

app/src/main/java/com/android/unio/ui/association/EditAssociation.kt
```diff
@@ -74,7 +74,6 @@ fun EditAssociationScreen(
             newAssociation,
             null,
             onSuccess = {
-              Log.d("EditAssociationScreen", "Association saved successfully.")
```
Zafouche, line 77: Good eye!

Zafouche (approved):
Great work on this! 🚀
No problems spotted on my side. (It's pretty much only documentation, so that's normal)

Much needed documentation work, this makes everything easily readable and understandable.

I only have one question but I think I answered my own question (since we had discussed how we would register users to an event) so I'll approve for sure anyway.

app/src/main/java/com/android/unio/model/event/Event.kt
Redd87: Yes precisely, I found it better to have a static max number of places and compute the remaining places as maxNumberOfPlaces - numberOfSaved

Zafouche:
Interesting bug, I could review event though it wasn't open for review (I thought it was because I saw a message saying it was opened). Turns out it wasn't, but it still allowed me to

Issue #271, opened 2024-12-08
Remove the "I'm going" button and add event highlighting on the address button in the event details page


Zafouche:
For the `I'm going` button, either: 
- Remove it, and make the address button perform a click on the event marker.
- Change it such that it's a button that sends the user on the Google Maps app (like on our app, when we click on the marker, it pops up two buttons at the bottom of the map, that serve as links to the default map apps of the phone, e.g. Google Maps for android)

Zafouche:
I ended up choosing the first option, as we can add an overlay button on the `GoogleMap` compose, that is essentially a link that takes the user to the Google maps app on their phone, and inputs the location query.

PR #272, opened 2024-12-08
Feat/event address button marker focus

Zafouche:
## Description

This PR addresses [this task](https://github.com/SwEnt-Group13/Unio/issues/271), an issue which was listed in a more general [bug summary issue](https://github.com/SwEnt-Group13/Unio/issues/269). It consists in removing the I'm going button from the event details page and improve the mechanism when pressing the event’s address button. The old approach simply centers the camera on the specific event’s location. This update improves the mechanism and highlights the marker on top of that, showing its info text, e.g. title and description.

Note: I will eventually add a second FAB button that shows up only when the map is opened from the event, and that is essentially a link to the Google Maps app of the phone, which opens it with the location query already filled out.

## Motivation and Context

Previously, the event details page had an I'm going button that was not used. This button had to be either used to take use to an assisted GPS, or removed. With @Redd87, we chose to remove it, because overlay buttons on the Google Maps (when clicking on a marker) shows buttons on the bottom of the screen which takes the user to the Google Maps app, with the marker's location as an input to the GPS.

By implementing the center-and-highlight logic on the map, the address button becomes more intuitive and the highlighted event is much more clear.

### MapViewModel.kt : 
- Added `_highlightedEventUid`: MutableStateFlow to store the UID of the event to highlight
- Added `setHighlightedEvent`: function that sets a highlighted event on the map given a uid and location, and stores its location.
- Added `clearHighlightedEvent`: function that clears the highlighted event's UID et location form the MutableStateFlows

### Map.kt :
- Added event highlighting when, for example, the user clicks on the address button in the event page. This buttons takes the user to the map, centers it on the marker and highlights it, meaning it shows its title and description.
- Moved the map center LaunchedEffect logic to  the EventMap composable.
- Added animations to all camera position updates

### EventDetails:
- Changed the address button (MAP_BUTTON test tag) to use the new `setHighlightedEvent` function

### MapViewModelTest: 
- Added test for `setHighlightedEvent`
- Added test for `clearHighlightedEvent `

### EventDetailsTest
- Added simple address button functionality to existing test, that checks the view model is updated correctly when clicking on it.


Zafouche:
The rest isn't testable without an e2e test (only the Map screen has uncovered lines in this PR), and I find it hardly worth it for this screen, given the time we have left (or even if we had more time).

PR #276, opened 2024-12-09
refactor(AssociationProfile): remove unusable features and add see le…

Romainhir:
…ss button for events

### DIFF
--- a/app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
+++ b/app/src/main/java/com/android/unio/ui/association/AssociationProfile.kt
@@ -20,10 +20,10 @@ import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material.icons.automirrored.filled.ArrowForward
 import androidx.compose.material.icons.automirrored.outlined.ArrowBack
 import androidx.compose.material.icons.filled.Add
-import androidx.compose.material.icons.outlined.Share
 import androidx.compose.material3.Button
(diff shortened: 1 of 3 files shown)


Romainhir:
Tell me if the cognitive complexity is really too high or if it is fine (i.e. readable). I can fix this if necessary but I do not think this is too complexe

Zafouche:
> Tell me if the cognitive complexity is really too high or if it is fine (i.e. readable). I can fix this if necessary but I do not think this is too complexe

I think its okay, however we should add comments to explain all the if else blocks (what they display namely).

Zafouche:
Actually, I do think reducing the complexity through factoring out some logic would be good. 

For example, for the Event creation button, you could factor out the logic that's inside the `if (isAdmin)` block, e.g. this part : 
```kotlin
Button(
        onClick = {
          if (isConnected) {
            navigationAction.navigateTo(Screen.EVENT_CREATION)
          } else {
            ToastUtils.showToast(context, context.getString(R.string.no_internet_connection))
          }
        },
        modifier = Modifier.testTag(AssociationProfileTestTags.ADD_EVENT_BUTTON),
        contentPadding = ButtonDefaults.ButtonWithIconContentPadding) {
          Icon(
              Icons.Filled.Add,
              contentDescription = context.getString(R.string.association_profile_add_event_button),
              modifier = Modifier.size(ButtonDefaults.IconSize))
          Spacer(Modifier.size(ButtonDefaults.IconSpacing))
          Text(context.getString(R.string.association_profile_add_event_button))
        }
``` 
which can be factored out into its own composable. This would greatly reduce cognitive complexity.

Romainhir:
I do not know why all these commits are registered on this branch. I may have made a manipulation error

Zafouche:
> I do not know why all these commits are registered on this branch. I may have made a manipulation error

Perhaps you’ve done a merge rebase or something as such by accident while updating from main?

PR #281, opened 2024-12-09
Fix/home sort event

Redd87:
This small PR aims to fix the issue " The events in Home seem to be sorted in a random order, or if not random, certainly aren't sorted in chronological order" brought in #269. It does so by sorting the events before passing them to the event list to be displayed. I originally also wanted to fix the weird interaction when you unsave an event in Saved.kt, but this seems to be a data synchronization issue between the repository and the view model, and will therefore not be done in this PR.


Romainhir (approved):
Nice small addition. LGTM

PR #287, opened 2024-12-09
Rework some functions/variables

Romainhir:
Some functions are currently only use in unit test and nowhere in the application. These functions should be removed. And some functions need to have some small change. Precisely, this means :

- methods getNextEventsFromDateToDate, getEventsOfAssociation, and getEventsForAssociation as well as their corresponding unit tests should be removed as they are unused
- serialization.kt's function mapUserToRoles and mapRolesToPermission should be private

I also add these weaknesses into this task :

- UnrememberedMutableState annotation in Saved should be checked and fixed.
- Hardcoded strings in timeUntilEvent in Map.kt

### DIFF
--- a/app/src/main/java/com/android/unio/ui/map/Map.kt
+++ b/app/src/main/java/com/android/unio/ui/map/Map.kt
@@ -1,6 +1,7 @@
 package com.android.unio.ui.map
 
 import android.Manifest
+import android.content.Context
 import android.content.pm.PackageManager
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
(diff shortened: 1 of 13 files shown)


Romainhir:
I don't know why I don't have the coverage required. I have not implemented any new features that required additionnal tests. Tell me if I can do something to improve it or if it is fine this way.

app/src/main/java/com/android/unio/ui/map/Map.kt
```diff
@@ -300,13 +302,32 @@ fun DisplayEventMarker(event: Event, customIconResId: Int?) {
...
+          context.getString(R.string.map_event_marker_time_remaining_days) +
+          hours +
+          context.getString(R.string.map_event_marker_time_remaining_hours))
+  else
+      (context.getString(R.string.map_event_marker_time_remaining_prefix) +
+          days +
+          context.getString(R.string.map_event_marker_time_remaining_days))
+}
+
+object MapCst {
```
Zafouche, line 325: Fair enough moving the constants to an object

app/src/main/java/com/android/unio/ui/saved/Saved.kt
```diff
@@ -60,20 +58,24 @@ fun SavedScreen(
...
-  val savedEvents by derivedStateOf {
-    allEvents
-        .filter { user!!.savedEvents.contains(it.uid) }
-        .sortedWith(compareBy({ it.startDate }, { it.uid }))
+  val savedEvents by remember {
+    derivedStateOf {
+      allEvents
+          .filter { user!!.savedEvents.contains(it.uid) }
+          .sortedWith(compareBy({ it.startDate }, { it.uid }))
+    }
```
Zafouche, line 66: Nice!

Zafouche (approved):
LGTM. Good cleanup!

Nothing to report really, so good job.

No concerns about the code coverage either, you didn't add any feature so it's fine.

PR #288, opened 2024-12-10
Rework Claim Association

Romainhir:
From #269. Some components in the "Claim Association" page behave not like expected. This task aims to correct this. Precisely, this means :

- The three dots top right do not do anything, and should be removed
- The search bar should be enabled by default, as it makes no sense for it to be toggled off. Or if for some reason it does make sense the logic for it should be implemented
- Create association" is currently unimplemented and should be removed

### DIFF
--- a/app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
+++ b/app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
@@ -41,7 +41,7 @@ fun AssociationRepositoryFirestore.Companion.serialize(association: Association)
  * @param members Members to map.
  * @return Map of user UIDs to role UIDs.
  */
-fun mapUsersToRoles(members: List<Member>): Map<String, String> {
+private fun mapUsersToRoles(members: List<Member>): Map<String, String> {
   return members.associate { member -> member.user.uid to member.role.uid }
 }
 
@@ -51,7 +51,7 @@ fun mapUsersToRoles(members: List<Member>): Map<String, String> {
  * @param roles Roles to map.
  * @return Map of role UIDs to role data.
  */
-fun mapRolesToPermission(roles: List<Role>): Map<String, Map<String, Any>> {
+private fun mapRolesToPermission(roles: List<Role>): Map<String, Map<String, Any>> {
   return roles.associate { role ->
     role.uid to
         mapOf(
--- a/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
+++ b/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
@@ -21,7 +21,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.focus.focusRequester
 import androidx.compose.ui.platform.LocalContext
-import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
 import com.android.unio.R
--- a/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
+++ b/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
@@ -50,7 +50,6 @@ fun UserClaimAssociationScreen(
   val context = LocalContext.current
 
   val focusRequester = remember { FocusRequester() }
-  val focusManager = LocalFocusManager.current
 
   LaunchedEffect(Unit) { focusRequester.requestFocus() }
 
--- a/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
+++ b/app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
@@ -5,7 +5,6 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.outlined.ArrowBack
-import androidx.compose.material3.Button
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/android/unio/model/firestore/transform/Serialization.kt
```diff
@@ -41,7 +41,7 @@ fun AssociationRepositoryFirestore.Companion.serialize(association: Association)
  * @param members Members to map.
  * @return Map of user UIDs to role UIDs.
  */
-fun mapUsersToRoles(members: List<Member>): Map<String, String> {
+private fun mapUsersToRoles(members: List<Member>): Map<String, String> {
```
Aurelien9Code, line 44: You may be right that the function should not be private here, however it means modifying the tests accordingly - as you did. For example in #262 I modify the Role object which means I should also remodify the tests ? Even if that's not too long, I think it's better to centralize everything that is just duplicated three times in the project. 
Wouldn't it be okay to put it in a tool file so that conceptually we can leave it `public` ?

app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
```diff
@@ -66,39 +72,31 @@ fun UserClaimAssociationScreen(
         ) {
           Column(
               modifier = Modifier.padding(16.dp),
-              verticalArrangement = Arrangement.spacedBy(6.dp),
+              verticalArrangement = Arrangement.spacedBy(12.dp),
               horizontalAlignment = Alignment.CenterHorizontally,
           ) {
             Text(
-                context.getString(R.string.user_claim_association_you_can_either),
+                context.getString(R.string.user_claim_association_you_can_do_the_following),
```
Aurelien9Code, line 79: If you don't leave a choice to the user in this page, maybe just remove this text as it gives no more information
It seemed weird when reading it in the screen

app/src/main/java/com/android/unio/ui/user/UserClaimAssociation.kt
```diff
@@ -66,39 +72,31 @@ fun UserClaimAssociationScreen(
...
-            Button(
-                onClick = {
-                  ToastUtils.showToast(
-                      context,
-                      context.getString(R.string.user_claim_association_not_implemented_yet))
-                },
-                modifier = Modifier.testTag(UserClaimAssociationTestTags.NEW_ASSOCIATION_BUTTON)) {
-                  Text(context.getString(R.string.user_claim_association_create_association))
-                }
+            val focusRequester = remember { FocusRequester() }
```
Aurelien9Code, line 86: It's a good idea to use the focusRequester ! I checked the doc about it and I think you are right to use it now ; )

app/src/main/java/com/android/unio/ui/user/UserClaimAssociationPresidentialRights.kt
```diff
@@ -106,22 +106,16 @@ fun UserClaimAssociationPresidentialRightsScreenScaffold(
...
-                      Icons.Outlined.MoreVert,
-                      contentDescription = context.getString(R.string.association_see_more))
-                }
-              }
             })
       },
       content = { padding ->
         Surface(
-            modifier = Modifier.padding(padding),
+            modifier = Modifier.fillMaxWidth().padding(padding),
```
Aurelien9Code, line 113: good catch

app/src/test/java/com/android/unio/model/firestore/HydrationAndSerializationTest.kt
```diff
@@ -116,8 +114,18 @@ class HydrationAndSerializationTest {
...
+        serialized["members"])
+    assertEquals(
+        association.roles.associate { role ->
+          role.uid to
+              mapOf(
+                  Role::displayName.name to role.displayName,
+                  Role::permissions.name to
+                      role.permissions.getGrantedPermissions().map { it.stringName })
+        },
+        serialized["roles"])
```
Aurelien9Code, line 128: I think it just make the tests less readable..

app/src/main/java/com/android/unio/ui/user/UserClaimAssociationPresidentialRights.kt
```diff
@@ -106,22 +106,16 @@ fun UserClaimAssociationPresidentialRightsScreenScaffold(
...
-            },
-            actions = {
-              Row {
-                IconButton(onClick = {}) {
-                  Icon(
-                      Icons.Outlined.MoreVert,
-                      contentDescription = context.getString(R.string.association_see_more))
-                }
-              }
             })
```
Aurelien9Code, line 109: Yes, I think I did not find anything to put in it in the end haha

Aurelien9Code (approved):
Hello ! 

You are correct to remove the three dots from the `UserClaimAssociationPresidentialRights` as there is nothing to put in it !
The search bar that is automatically toggle on is a good choice !
However, why remove the Create Association button ? We let multiple times in the project not-yet-implemented `@Composable`, given that we show it with a Toast. There was a point for the AppSearchBar not to be toggled on - it was because there was another choice on the page : creating an association. 

But even if I can disagree with the justification as it may just make the next developper that want to implement the creating association logic loose time (Redo the button, connect it to the saveAssociation function, rewrite the text "You can either...", re let the AppSearchBar not be toggled on automatically...), I will still approve it.

The reason is that in the end, when checking out on your PR and looking at the screen, it may be better to put the "create association" and "claim association" logics into the three dots of the UserProfile. Then as the choice is made before the AppSearchBar can be automatically toggle on with the FocusRequester (amazing btw), there is no need for the Text to be there anymore, and we fall back to everything you did.

No need to do it here, I can refactor it when your PR is merged. 

Can you just remove the unuseful Text "user_claim_association_you_can_do_the_following" ? And maybe let the `mapUsersToRoles` and `mapRolesToPermission` publics but in a tool file, except if you see a real problem about it. It just allows not to loose a whole `connectCheck` time because the tests are not changed when these functions are.

PR #290, opened 2024-12-10
Polish Association profile

Romainhir:
The issue https://github.com/SwEnt-Group13/Unio/issues/269 list many bug/flaw in our application. In this task, I'm going to fix all the issue mentionned there that regard the association profile. This means :

- The Share button is currently not implemented
- Unimplemented job offers in Association profile
- See more" is displayed even when all the events of the association are already displayed; should switch to "See less" and hide events if clicked again

Note : this is the second version of this PR. The previous one has been closed because a (skill) issue occured. See #276


armouldr (approved):
Nice fix to the see more button that was a bit broken, it looks good to me now :rocket: 
I approve the PR as I don't see any important problem, but please read my comments before merging :muscle:

Issue #251, opened 2024-12-05
Navigation from welcome Screen to Account Details


AlouchLaBouche:
I noticed that if we create an account in welcome screen and verify it via the sent email, and then fore some reason quit the application. When we re input our email and password, a toast appears signaling us that we have indeed signed in, but no navigation is performed towards the `AccountDetails.kt` Screen.
This is an issue that must be worked on, as blocking on the welcome screen should not occur !

AlouchLaBouche:
I've found one of the issues. It stems from the `getUserWithId`method that is in userRepository.kt. Basically we try to fetch a user in Firestore but he hasn't filled out his account details yet (he doesn't exist in the Firestore DB), therefore we get an error : `document not found` but we don't handle this error correctly as it is "to be expected" to some extent as we should redirect the user to fill in his AccountDetails.

I've added some error handling in the method that throws the correct error in the onFailure callback, thus we do get redirected to the accountDetails page.

Sadly another error occurs when verifying ones account. Same principle, get to the email verification screen, leave the app, come back, verify your email, click on refresh, and nothing happens.... If we reload the app again we end up in AccountDetails as the email has been modified in the DB, therefore the error is client side. Further investigation is necessary

AlouchLaBouche:
After some investigation it seems that the error is much more complex than originally thought. The navigation from email verification screen to account details didn't work because of `Firebase.auth.currentUser?.reauthenticate(userViewModel.credential!!)`. And it makes perfect sense as when you create your account you go through the welcome screen and therefore you input some credentials (email, password). But if you leave the app in the mean time, and re open it, it will redirect you straight to the email verification screen without going through the welcome srceen due to the behavior of `addAuthStateVerifier()` in the Auth view model. This means that the crendentials will be null and that the request will fail and we stay blocked in the email verification screen... 
This can be fixed with by using some other method such as `Firebase.auth.currentUser.getIdToken(true)` that will force to get the user without the need to store the credentials.  

But another issue occurs and it is linked to the security rules, therefore it only concerns the emulated data base as the security rules are not present on the production one. Basically, whatever method we use to authenticate when verifying one's email, be it getIdToken or reauthenticate, the user must sign out and sign in for the security rules to register and realize that the user's email is indeed verified (or wait 1h). Therefore, when first logging in, all the requests of all the view model fail due to "permission denied".
This is, how you can imagine, far from ideal and there are two solutions to this I believe : 
[...]

PR #292, opened 2024-12-11
Fix  [Issue 251] Navigation during the login process

AlouchLaBouche:
This PR aims to resolve the issue mentioned [here](https://github.com/SwEnt-Group13/Unio/issues/251). The main issues that have been addressed are edge cases during the login process of the application, typically users that close the application between the navigation from the Welcome Screen to the Email Verification Screen and also to the Account Creation screen.
Another major issue that has been fixed is the correct handling of the repositories with regards to the Firebase security rules. Basically all are requests fail on the emulated data base (where the Firebase security rules are present) when first logging in, because the repositories fetch the data before the user has the chance to verify his email (they fetch the data on the creation of the user, and not on his verification), therefore all the requests fail.

To fix these issues I've changed a couple of things:
1. The credentials have been moved from `userViewModel` to `authViewModel` as it made more sens that this field is handled in the `authViewModel`.
2. I've added some checks to all the repositories so that the user must have his email verified before fetching the data. This is to avoid queries that will fail due to the Firebase rules that are in place.
3. I've added a callback in the email verification composable that requests all the repositories to fetch the data. This is done in the `onEmailVerified` Callback
4. Some refactoring has been made to the `addAuthStateVerifier()` to better ensure navigation between composables during the login process, noticeably we ensure that the user has filled out his credentials before verifying his email.
5. If the user's email is not verified in the welcome screen, we simply navigate to the email verification screen manually.


app/src/androidTest/java/com/android/unio/components/authentication/WelcomeTest.kt
```diff
@@ -34,19 +41,20 @@ class WelcomeTest : TearDown() {
...
+  @MockK private lateinit var firebaseAuth: FirebaseAuth
 
   @Before
   fun setUp() {
     MockKAnnotations.init(this)
 
-    //    mockkStatic(FirebaseAuth::class)
-    //    every { Firebase.auth } returns firebaseAuth
-    //    every { firebaseAuth.addAuthStateListener(any()) } just runs
-    //    every { firebaseAuth.removeAuthStateListener(any()) } just runs
```
Zafouche, line 49: Nice catch, seems like commented code found its way on the main branch...

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
```diff
@@ -31,6 +32,19 @@ constructor(private val firebaseAuth: FirebaseAuth, private val userRepository:
...
+
+  /**
+   * Sets the [AuthCredential] object to be used for linking accounts.
+   *
+   * @param credential The [AuthCredential] object to set.
+   */
+  fun setCredential(credential: AuthCredential?) {
+    _credential = credential
+  }
+
```
Redd87, line 47: Rather than having for _crediential a getter and setter that does not impose any constraints on the credential param, couldn't we have _credential public?

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
```diff
@@ -65,20 +79,21 @@ constructor(private val firebaseAuth: FirebaseAuth, private val userRepository:
...
-                Log.e("UnioApp", "Error fetching account details: $it")
-                _authState.value = Screen.WELCOME
+              onSuccess = { _authState.value = Screen.HOME },
+              onFailure = { error ->
+                if (error is FirebaseFirestoreException &&
+                    error.code == FirebaseFirestoreException.Code.NOT_FOUND) {
+                  _authState.value = Screen.ACCOUNT_DETAILS
+                } else {
+                  _authState.value = Screen.WELCOME
+                }
```
Zafouche, line 89: Nicely done! This looks much better and fixes our edge case bug!

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
```diff
@@ -65,20 +79,21 @@ constructor(private val firebaseAuth: FirebaseAuth, private val userRepository:
...
               })
         } else {
-          _authState.value = Screen.EMAIL_VERIFICATION
+          if (_credential == null) {
+            _authState.value = Route.AUTH
+          } else {
```
Zafouche, line 94: It is good to redirect the user to `Auth` if something goes wrong 👍

app/src/main/java/com/android/unio/model/user/UserViewModel.kt
Zafouche, line 1: It's good that you extracted out the logic to the Auth view model

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
```diff
@@ -65,20 +79,21 @@ constructor(private val firebaseAuth: FirebaseAuth, private val userRepository:
...
-              {
-                Log.e("UnioApp", "Error fetching account details: $it")
-                _authState.value = Screen.WELCOME
+              onSuccess = { _authState.value = Screen.HOME },
+              onFailure = { error ->
+                if (error is FirebaseFirestoreException &&
+                    error.code == FirebaseFirestoreException.Code.NOT_FOUND) {
+                  _authState.value = Screen.ACCOUNT_DETAILS
+                } else {
+                  _authState.value = Screen.WELCOME
```
Redd87, line 88: Comment this as we've discussed the logic of checking type of error, but for future reference knowing that this is if the user only partially initialized his account will be useful

app/src/main/java/com/android/unio/model/user/UserViewModel.kt
```diff
@@ -44,10 +43,6 @@ constructor(
   private var updateJob: Job? = null
   private var initializeWithAuthenticatedUser: Boolean = true
 
-  private var _credential: AuthCredential? = null
```
Redd87, line 47: I agree with moving this to the authviewmodel, it's cleaner this way

app/src/main/java/com/android/unio/model/user/UserRepositoryFirestore.kt
```diff
@@ -18,7 +19,7 @@ class UserRepositoryFirestore @Inject constructor(private val db: FirebaseFirest
 
   override fun init(onSuccess: () -> Unit) {
     Firebase.auth.registerAuthStateListener {
-      if (it.currentUser != null) {
+      if (it.currentUser != null && it.currentUser!!.isEmailVerified) {
```
Zafouche, line 22: The part that checks if the user is null becomes useless (e.g. `it.currentUser != null`) with this addition. Redundant null check, you should remove this.

app/src/main/java/com/android/unio/ui/authentication/EmailVerification.kt
```diff
@@ -48,7 +49,11 @@ import com.google.firebase.auth.auth
  */
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun EmailVerificationScreen(navigationAction: NavigationAction, userViewModel: UserViewModel) {
+fun EmailVerificationScreen(
+    navigationAction: NavigationAction,
+    authViewModel: AuthViewModel,
+    onEmailVerified: () -> Unit
```
Redd87, line 55: Modify the javadoc to reflect the signature change

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
```diff
@@ -64,7 +65,7 @@ import com.google.firebase.auth.auth
  * @param userViewModel The [UserViewModel] to use.
```
Redd87, line 65: Same here, update the javadoc

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
```diff
@@ -186,7 +187,12 @@ fun WelcomeScreen(navigationAction: NavigationAction, userViewModel: UserViewMod
  * @param password The password to use.
  * @param context The context to use.
  */
-fun handleAuthentication(email: String, password: String, context: Context) {
+fun handleAuthentication(
+    email: String,
+    password: String,
+    context: Context,
+    navigationAction: NavigationAction
```
Redd87, line 194: And here too

app/src/main/java/com/android/unio/ui/authentication/Welcome.kt
```diff
@@ -209,16 +215,7 @@ fun handleAuthentication(email: String, password: String, context: Context) {
...
-          signInResult.user.sendEmailVerification().addOnCompleteListener {
-            if (it.isSuccessful) {
-              ToastUtils.showToast(
-                  context, context.getString(R.string.welcome_toast_verification_sent))
-            } else {
-              ToastUtils.showToast(
-                  context, context.getString(R.string.welcome_toast_verification_sent_failed))
-              Log.e("WelcomeScreen", "Failed to send verification email", it.exception)
-            }
-          }
```
Zafouche, line 221: Fair deletion 👍 Since we now redirect the user to the `Auth` screen if the user closes the app when they have to verify their email, this edge case disappears.

Redd87 (changes requested):
Good changes that fix previously overlooked vulnerabilities, well done! A few comments to update and I have a question about the _credential variable

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
```diff
@@ -24,7 +24,7 @@ class AssociationRepositoryFirestore @Inject constructor(private val db: Firebas
 
   override fun init(onSuccess: () -> Unit) {
     Firebase.auth.registerAuthStateListener {
-      if (it.currentUser != null) {
+      if (it.currentUser != null && it.currentUser!!.isEmailVerified) {
```
Zafouche, line 27: Redundant null check

app/src/main/java/com/android/unio/model/search/SearchRepository.kt
```diff
@@ -54,7 +54,7 @@ constructor(
   suspend fun init() {
     withContext(Dispatchers.IO) {
       Firebase.auth.registerAuthStateListener {
-        if (it.currentUser != null) {
+        if (it.currentUser != null && it.currentUser!!.isEmailVerified) {
```
Zafouche, line 57: Redundant null check

Zafouche (changes requested):
Great work! 🚀

Not much to say, however I do request small changes that should be done in the repositories, e.g. redundant null checks. Check my comments under.

app/src/main/java/com/android/unio/model/authentication/AuthViewModel.kt
AlouchLaBouche: Yes you are right, it's a pretty messy handling of navigation.

app/src/main/java/com/android/unio/ui/authentication/EmailVerification.kt
AlouchLaBouche: Good catch ! You are right we often forget to update the java doc and forget what we have done.

app/src/main/java/com/android/unio/model/association/AssociationRepositoryFirestore.kt
AlouchLaBouche: My bad, I've told you this but actually these checks are indeed very useful. If they are not in place, null assertion errors get thrown. 
We don't want to assert that the user is null as he can totally be null (for example when he doesn't have an account yet). Nevertheless the onSuccess method should only be called when the user is not null that is checked first, and then that his email is verified (Recall Lazy checks in software construction).

Redd87 (approved):
LGTM

PR #293, opened 2024-12-11
Feat/ Make sure that all related images are deleted when deleting a data object.

armouldr:
For now, image deletion logic is only implemented for the user profile pictures, meaning that event and association images can be uploaded but cannot be deleted in another way than manually taking them out of Firebase. This is bad, as it will lead us to flood our storage.

Now when deleting an event : 
-Event banner (main image) get deleted from Firebase Storage
-Event user pictures get deleted from Firebase Storage
-Event user pictures get deleted from Firestore

The low code coverage is due to onFailure callbacks not being tested, but I don't think we tested them anywhere as it doesn't bring much except coverage.


Aurelien9Code (changes requested):
/Very good implementation ! It is very important that the database should not be fullfilled with random unused data. Also, your implementation is perfect, I looked into an tried every your new functions in Android studio and everything seems to work perfect ! Consider add a few comments as some of our friends lost much time to do it for the whole project. Thought, this is really a very useful PR 🚀

However when checking out in your PR and using your delete event button in main, I first thought that it did not work as the `EventCard` in the `AssociationEventCard` is not updated directly. 

I know this is none of your current code, but for this kind of PR, when showing your work to the assistant tomorrow we need to directly see it. Consider fixing this bug so that we can see the work of your PR directly in the UI after clicking on the delete event button.

Here's an idea of implementation : 
First your `deleteEvent` function should automatically update the `associationViewModel` and not rely on a `fetchAll()` of the database to modify the UI. As you can not directly do it in the `EventViewModel`, you can do it in the success callback of `deleteEvent()` in the `EventEdit` screen. As this is only a call to a viewModel function (just as the call to `deleteEvent`), this seems fair to do it here. 

Now you need the viewModel function to update the events field of the current selected association : `/**
     * Deletes an event from the events list of the selected association locally.
     *
     * @param eventId The ID of the event to be deleted.
     */
    fun deleteEventLocally(eventId: String) {
        val selectedAssociation = _selectedAssociation.value

        if (selectedAssociation != null) {
            val eventToDelete = selectedAssociation.events.uids.find { it == eventId }

            // if event exists ->  remove it from the events list
            if (eventToDelete != null) {
                selectedAssociation.events.remove(eventId) //check the definition of remove to see that it does not fetch the database ; )
                
                val updatedAssociation = selectedAssociation.copy(events = selectedAssociation.events)
                
                _selectedAssociation.value = updatedAssociation
                _associations.value = _associations.value.map { // update global list of associations
                    if (it.uid == selectedAssociation.uid) updatedAssociation else it
                }
            } else {
                Log.e("AssociationViewModel", "Event with ID $eventId not found")
            }
        } else {
            Log.e("AssociationViewModel", "No association selected to delete event from")
        }
    }`
    
then you could modify your `onSucess` callback implementation by 

`eventViewModel.deleteEvent(
                          eventToEdit,
                          onSuccess = { associationViewModel.deleteEventLocally(eventToEdit.uid); navigationAction.goBack() },
                          onFailure = {
                            Toast.makeText(
                                    context,
                                    context.getString(R.string.event_edit_failed),
                                    Toast.LENGTH_SHORT)
                                .show()
                          })` in the `EventEdit` screen.
(As my emulator crashes indefinitely, I could just build this structure in Android studio but not verify it thought)
                      
Adding some of thoses lines and testing them (maybe already done) could be a good way to improve your coverage to +70% ;)

app/src/main/java/com/android/unio/model/strings/Strings.kt
```diff
@@ -9,8 +9,9 @@ object FirestorePathsStrings {
 
 object StoragePathsStrings {
   const val USER_IMAGES = "images/users/"
-  const val ASSOCIATION_IMAGES = "images/associations"
-  const val EVENT_PICTURES = "images/events/userPictures/"
+  const val ASSOCIATION_IMAGES = "images/associations/"
+  const val EVENT_IMAGES = "images/events/"
+  const val EVENT_USER_PICTURES = "images/events/userPictures/"
```
Aurelien9Code, line 14: It's more clear now

app/src/test/java/com/android/unio/model/event/EventUserPictureRepositoryFirestoreTest.kt
```diff
@@ -64,4 +64,13 @@ class EventUserPictureRepositoryFirestoreTest {
...
   }
+
+  @Test
+  fun testDeleteEventById() {
+    `when`(collectionReference.document(eventUserPicture.uid)).thenReturn(documentReference)
+    `when`(voidTask.addOnSuccessListener(any())).thenReturn(voidTask)
+    `when`(documentReference.delete()).thenReturn(voidTask)
+    eventUserPictureRepository.deleteEventUserPictureById(
+        eventUserPicture.uid, {}, { e -> throw e })
+  }
```
Aurelien9Code, line 75: Difficult mocks ! Well done !

Aurelien9Code:
[UPDATE] As you remind me, this was not due to your PR, as you concentrate on deleting the image and not the whole event. When trying to delete an event UI side and then check if the image was effectively deleted, I found this problem and forgot it was only about images. 

I finally made the implementation that I started proposing you  work, so when you will merge this PR I'll quickly copy it in a refactor. Sorry for this. 

I don't think -70% line coverage is okay usually - even for callbacks, but as this is a small PR this should not affect main too much. 
Then, I would say your PR is perfect. Well done 🚀

Aurelien9Code (approved):
[UPDATE] As you remind me, this was not due to your PR, as you concentrate on deleting the image and not the whole event. When trying to delete an event UI side and then check if the image was effectively deleted, I found this problem and forgot it was only about images.

I finally made the implementation that I started proposing you work, so when you will merge this PR I'll quickly copy it in a refactor. Sorry for this.

I don't think -70% line coverage is okay usually - even for callbacks, but as this is a small PR this should not affect main too much.
Then, I would say your PR is perfect. Well done 🚀

armouldr:
Thanks for the review ! The only way to increase the code coverage would be to write some tests that execute the onFailure() callbacks of the methods in EventViewModel. Given that they only call Logs, the only purpose of this would be to higher our code coverage without testing anything.

PR #297, opened 2024-12-12
Pull to refresh indicators

oskar-codes:
This pull request adds pull to refresh indicators just like the one in the user's profile in the following places:
- Association Profile
- Event Details
- Home Screen
- Saved Screen
- Explore Screen

This pull request also **temporarily** fixes the issue where a newly created event would not show up on the association's page. This fix should be refactored soon as it produces excessive requests. EDIT: This has been fixed.

PR #298, opened 2024-12-13
Fix/add delete event local

Aurelien9Code:
This PR fixes a bug that is that the just added event and the just deleted event are not immediately visible in the UI. It should be coupled with a force fetch that @oskar-codes did in #297. 

But as this is done here, it saves us from doing an unnecessary request to the db, as we already have the event we push to the db (no need to fetch it again) and it saves us from fetching all the event list to delete an event (we call also do it locally).

### DIFF
--- a/app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
+++ b/app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
@@ -2,6 +2,7 @@ package com.android.unio.model.association
 
 import android.util.Log
 import androidx.lifecycle.ViewModel
+import com.android.unio.model.event.Event
 import com.android.unio.model.event.EventRepository
 import com.android.unio.model.follow.ConcurrentAssociationUserRepository
 import com.android.unio.model.image.ImageRepository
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
```diff
@@ -133,6 +134,48 @@ constructor(
...
+   * @param eventId The ID of the event to be deleted.
+   */
+  fun deleteEventLocally(eventId: String) {
+    val selectedAssociation = _selectedAssociation.value
+    if (selectedAssociation != null) {
+      val eventToDelete = selectedAssociation.events.uids.find { it == eventId }
+      // if event exists ->  remove it from the events list
+      if (eventToDelete != null) {
+        selectedAssociation.events.remove(
+            eventId) // check the definition of remove to see that it does not fetch the database ;
```
armouldr, line 149: I think the comment about checking the definition was meant to be temporary

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
```diff
@@ -75,7 +86,9 @@ class FirestoreReferenceList<T : UniquelyIdentifiable>(
    * @param uid The UID to remove.
    */
   override fun remove(uid: String) {
-    _uids.remove(uid)
+    if (_uids.remove(uid)) {
```
armouldr, line 89: I didn't know about List.remove returning a boolean that represents whether it made a change. Avoids an unnecessary operation :)

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -306,8 +305,14 @@ fun EventCreationScreen(
...
                         eventPictures = MockReferenceList(),
-                    ),
-                    onSuccess = { navigationAction.goBack() },
+                    )
+                eventViewModel.addEvent(
+                    inputStream,
+                    newEvent,
+                    onSuccess = {
+                      associationViewModel.addEventLocally(newEvent)
+                      navigationAction.goBack()
```
armouldr, line 314: Nice again, makes the app appear "refreshed" without requesting data it already has

app/src/main/java/com/android/unio/ui/event/EventEdit.kt
```diff
@@ -244,7 +244,10 @@ fun EventEditScreen(
                       // A dialog should be added to prevent accidental deletion
                       eventViewModel.deleteEvent(
                           eventToEdit,
-                          onSuccess = { navigationAction.goBack() },
+                          onSuccess = {
+                            associationViewModel.deleteEventLocally(eventToEdit.uid)
```
armouldr, line 248: same

armouldr (commented):
Nice work ! This is going to make the app look way better and responsive. Just make sure to make the tiny forgotten comments/docs and I will approve :)

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
Aurelien9Code: It was for my reviewers - meaning it was for you ;) 
I'll remove it

Aurelien9Code: Update, I still think it makes it clear for the people reading it. Can I keep it ?

app/src/main/java/com/android/unio/model/firestore/FirestoreReferenceList.kt
Aurelien9Code: Yeah that's crazy !

app/src/main/java/com/android/unio/model/association/AssociationViewModel.kt
armouldr: Then maybe shorten it to `// does not fetch the database`, it's true that it could clarify the code a bit

AlouchLaBouche (approved):
This is some great fixes you have done here, well done :rocket:  ! 
The coverage leaves to be desired but as this is a small PR, It's not too big of a deal.
Great work !

PR #304, opened 2024-12-15
Feature: maximise event picture when clicking on it

armouldr:
The goal of this PR is to make the user pictures gallery more user-friendly and interactive by allowing user to click on a picture in order to view it in full screen. Additionnaly, the user can navigate between the full-screen pictures by scrolling horizontally or clicking on the right/left arrow buttons.


Aurelien9Code:
From 7.14% to 87.1%, bravo ! 🚀

armouldr:
Road to 1000% coverage :rocket: :sunglasses:

app/src/androidTest/java/com/android/unio/components/event/EventDetailsTest.kt
```diff
@@ -81,8 +82,16 @@ class EventDetailsTest : TearDown() {
     MockKAnnotations.init(this, relaxed = true)
     eventPictures =
         listOf(
-            EventUserPicture("12", "http:image.com", User.emptyFirestoreReferenceElement(), 0),
-            EventUserPicture("34", "http:image.com", User.emptyFirestoreReferenceElement(), 3))
+            EventUserPicture(
+                "12",
+                "https://i1.sndcdn.com/artworks-000055240636-kowvdx-t500x500.jpg",
```
Romainhir, line 87: What happend if the URL is not available (like a 404 error) ? If it does not make the tests crash, you can leave it as this, but otherwise consider downloading the image or use one already downloaded

app/src/androidTest/java/com/android/unio/end2end/EventCreationE2ETest.kt
```diff
@@ -275,7 +275,7 @@ class EventCreationE2ETest : EndToEndTest() {
     composeTestRule.waitUntilExactlyOneExists(
         matcher =
             hasTestTag(EventCreationTestTags.LOCATION_SUGGESTION_ITEM_LATITUDE + EVENT_LATITUDE),
-        timeoutMillis = 10000)
+        timeoutMillis = 15000)
```
Romainhir, line 278: Waiting 15 seconds befor etimeout is a lot. Confirm me you really need to wait that long, otherwise, reduce the sleeping time

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
```diff
@@ -0,0 +1,113 @@
...
+  val context = LocalContext.current
+  val iconSize = 40.dp
+  val onClickArrow = { isRight: Boolean ->
+    if (isRight && pagerState.currentPage < eventPictures.size - 1) {
+      scope.launch { pagerState.animateScrollToPage(pagerState.currentPage + 1) }
+    } else if (!isRight) {
+      if (pagerState.currentPage > 0) {
+        scope.launch { pagerState.animateScrollToPage(pagerState.currentPage - 1) }
+      } else {}
+    } else {}
```
Romainhir, line 58: Empty else statement. Consider handle these cases properly

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
```diff
@@ -0,0 +1,113 @@
...
+    eventPictures: List<EventUserPicture>
+) {
+  val scope = rememberCoroutineScope()
+  val context = LocalContext.current
+  val iconSize = 40.dp
+  val onClickArrow = { isRight: Boolean ->
+    if (isRight && pagerState.currentPage < eventPictures.size - 1) {
+      scope.launch { pagerState.animateScrollToPage(pagerState.currentPage + 1) }
+    } else if (!isRight) {
+      if (pagerState.currentPage > 0) {
```
Romainhir, line 55: Why did you make two if statement ? Why did you not put it in a single one like you do just before (with an AND operator) ?

Romainhir (changes requested):
Nice job, a few small details to clarify and I'll approve it

app/src/androidTest/java/com/android/unio/components/event/EventDetailsTest.kt
armouldr: Yes you are right, it made the test fail. I found a way to get the local Uri of a drawable instead.

app/src/androidTest/java/com/android/unio/end2end/EventCreationE2ETest.kt
armouldr: Nice spot, I had a problem with the test and wanted to see if it was due to insufficient delay, but I forgot to reset the correct timeout

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
armouldr: I did this because android studio was giving me the warning "'if' must have both main and 'else' branches if used as an expression" when I didn't put any else statement. But found that casting onClickArrow to (Boolean) -> Unit achieves the same and is cleaner than empty else statements :)

Romainhir (approved):

PR #307, opened 2024-12-15
Chore/figma adaptations

Zafouche:
## Description

The purpose of this PR is to fix the small inconsistencies between the `Figma` and the actual implementation.
Some fields or icons have been forgotten to be implemented, while for others, we had to take different paths so the changes had to be done in the `Figma` (obviously not visible in this PR).

## Motivation and Context
The reason behind this PR is simply to improve consistency between our design and our implementation, and make sure we haven't forgotten to implement anything what was designed.

### EventCreation
- Added `clear` trailing icon button to event title input field.
- Added `clear` trailing icon button to event short description field.

### EventEdit:
- Added `clear` trailing icon button to event title input field.
- Added `clear` trailing icon button to event short description field.

### EventEditComponents
- Added simple location icon to location input field.

### AccountDetails
- Added `clear` trailing icon button to user first name input field.
- Added `clear` trailing icon button to user last name input field.

### UserProfileEdition
- Added `clear` trailing icon button to user first name input field.
- Added `clear` trailing icon button to user last name input field.

## How has this been tested?
- EventCreation: simple tests that check the clear buttons' functionalities. 
- EventEdition: same as above
- AccountDetails: same
- UserEdition: same

## What's left:
Other screens need to be checked with the `Figma`, will be handled in different PRs (@Redd87 has already started a PR for other screens for instance).


app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
```diff
@@ -340,7 +355,20 @@ private fun UserTextFields(
           onLastNameChange(it)
         }
       },
-      value = lastName)
+      value = lastName,
+      trailingIcon = {
+        if (lastName.isNotEmpty()) {
```
Redd87, line 360: Instead of hiding the button if there is no text, consider maybe simply disabling the button, that way there is no big UI difference

app/src/main/java/com/android/unio/ui/event/EventCreation.kt
```diff
@@ -145,6 +147,19 @@ fun EventCreationScreen(
                                 Modifier.testTag(EventCreationTestTags.TITLE_CHARACTER_COUNTER))
                       }
                     }
+              },
+              trailingIcon = {
```
Redd87, line 151: Same here

Redd87 (approved):
Well done! the app is getting more polished everyday. I just had a comment about if we should hide the IconButton but it is debatable, so let me know what you think

app/src/main/java/com/android/unio/ui/authentication/AccountDetails.kt
Zafouche: Yeah I agree, it is best to let the user know the feature (button) exists before they event type anything. I'll do that before merging.

PR #314, opened 2024-12-17
Add highlighting to missing inputs in Event creation and Event edition

Romainhir:
This issue was flagged https://github.com/SwEnt-Group13/Unio/issues/269. It is part of polishing the inconsistencies of our app. In event creation/edition, you cannot save your work if some inputs are not filled out, but currently, it is not shown which input is missing. This task aims solve this issue


Romainhir:
Oh you are right, I forgot it, my bad

Zafouche (approved):
No problem.

Just pulled the change and tested it, it works. All done!

PR #315, opened 2024-12-17
Feature: Allow users to delete and like event pictures

armouldr:
From another PR of mine, you can now click on user pictures from the EventDetails. But this menu still lacks responsivity and feels very static. That's why I am adding the possibility to like and delete a picture :)

When you click on a picture, you can see:
-> Icon to like/dislike a picture
-> Counter of likes
-> the author's icon and name
-> a delete button


app/src/main/java/com/android/unio/model/event/EventViewModel.kt
```diff
@@ -212,25 +212,32 @@ constructor(
...
-                it,
-                {},
-                { e ->
-                  Log.e("EventViewModel", "An error occurred while loading associations: $e")
-                })
-            it.events.requestAll()
-          }
-        },
-        lazy = true)
+    if (updateAssociation) {
```
Aurelien9Code, line 223: Very good idea to put the boolean, it allows you not to update all the association as you don't need to here

app/src/main/java/com/android/unio/model/event/EventViewModel.kt
```diff
@@ -317,21 +326,86 @@ constructor(
...
+      event: Event,
+      onSuccess: () -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+
+    eventUserPictureRepository.deleteEventUserPictureById(
+        uid,
+        {
+          event.eventPictures.remove(uid)
+          updateEventWithoutImage(
```
Aurelien9Code, line 393: Instead of doing this - that will in firebase for every field to change, you could have just override the image field to "" to avoid unnecessary server computing time

app/src/main/java/com/android/unio/ui/event/EventDetails.kt
```diff
@@ -479,7 +499,10 @@ fun EventDetailsPicturesTab(event: Event, context: Context) {
                         .clip(RoundedCornerShape(10))
                         .testTag(EventDetailsTestTags.USER_EVENT_PICTURE + item.uid)
                         .clickable {
-                          scope.launch { pagerState.scrollToPage(index) }
+                          scope.launch {
+                            eventPictures.forEach { pic -> pic.author.fetch(lazy = true) }
+                            pagerState.scrollToPage(index)
```
Aurelien9Code, line 504: This seems a bit too much a fetching right here, but as this is not a big amount of data this is really fine

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
```diff
@@ -43,18 +67,68 @@ import kotlinx.coroutines.launch
...
   val scope = rememberCoroutineScope()
   val context = LocalContext.current
-  val iconSize = 40.dp
+  val arrowSize = 40.dp
+  var enableButton by remember { mutableStateOf(true) }
+  val event by eventViewModel.selectedEvent.collectAsState()
+
+  if (event == null) {
+    Log.e("PictureOverlay", "Event is null")
+    Toast.makeText(LocalContext.current, "An error occurred.", Toast.LENGTH_SHORT).show()
```
Aurelien9Code, line 83: Consider putting the good error in the Toast if you have it, here : "Event is null"

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
```diff
@@ -84,12 +187,11 @@ fun PictureOverlay(
...
-                          containerColor = MaterialTheme.colorScheme.primary)) {
+                          contentColor = MaterialTheme.colorScheme.onPrimary)) {
                     Icon(
                         Icons.AutoMirrored.Filled.ArrowBack,
                         context.getString(R.string.event_details_content_description_arrow_left),
-                        modifier = Modifier.size(iconSize))
+                        modifier = Modifier.size(arrowSize))
                   }
 
               IconButton(
```
Aurelien9Code, line 197: you could remove the arrow right if we are on the lastPage and remove the arrow left if we are on the first page, this should be done very easily with the pageState like `if pagerState.currentPageOffsetFraction == -0.5` for example

app/src/main/java/com/android/unio/ui/event/overlay/PictureOverlay.kt
```diff
@@ -99,13 +201,80 @@ fun PictureOverlay(
...
+                  horizontalArrangement = Arrangement.SpaceBetween,
+                  verticalAlignment = Alignment.CenterVertically) {
+                    Row(verticalAlignment = Alignment.CenterVertically) {
+                      IconButton(
+                          onClick = onClickLike,
+                          modifier =
+                              Modifier.testTag(EventDetailsTestTags.EVENT_PICTURES_LIKE_BUTTON),
+                          colors =
+                              IconButtonDefaults.iconButtonColors(
+                                  contentColor = MaterialTheme.colorScheme.onPrimary)) {
```
Aurelien9Code, line 226: Consider hardcoding the colors because as you said, it seems better when being in light mode

firebase/tests/firestore-rules.test.js
Aurelien9Code, line 1: So much valuable insights !

firestore.rules
Aurelien9Code, line 1: As I looked a bit the changes you made all the day and your explanations, I don't think there's anything to change here

Aurelien9Code (approved):
Very good PR ! Looks good to me ! 
If you don't have time to do the changes, I think it's okay as they are all minors

firestore.rules
armouldr: Thank you for your review !

PR #316, opened 2024-12-18
Style/profile edition UI

Redd87:
This short PR aims to fix some of the UI inconsistencies noted in #269, notably the padding in the flowRows of the profile and the size of the text fields.


oskar-codes (approved):
Absolutely outstanding work! The implementation is flawless and the attention to detail is impeccable. This is a phenomenal contribution—brilliant job!

PR #317, opened 2024-12-18
Fix/firestore security rules

Redd87:
This small PR updates the firestore security rules and opens it for review before I deploy them on Firestore. They aim to fix the permissions problems caused by outdated rules regarding properties for the Event and Association classes


Redd87:
As @armouldr already took care of the security rules, I'm closing this PR as it doesn't serve any purpose anymore to merge it

PR #318, opened 2024-12-18
Feature [Event Creation/Edition] Add possibility to add types to events

AlouchLaBouche:
This PR adds the possibility to add types to events ! This can be done during the Event creation process or when you edit an Event. This feature is reminiscent of the user interests, as they serve practically the same purpose, but for two different objects. 
This Feature is especially useful as it is necessary to add event types for the custom recommendation algorithm to work (Which will function on the basis of matching interests and types to compute a score for each event in regard to each user and then sorting the list of events).

Therefore this PR adds:
- A couple of new Interests for the user to choose from.
- Multiple new Event Types.
- The possibility to add/modify event types for each event (this was previously impossible).
- UI tests to ensure functionality.


app/src/main/java/com/android/unio/model/event/Event.kt
```diff
@@ -81,15 +75,33 @@ data class Event(
...
+      EventColors.TechPresentation, R.string.event_type_tech_presentation), // + Tech, Science
+  SCIENCE_FARE(EventColors.ScienceFare, R.string.event_type_science_fare), // + Science, Tech
+  FOOD_DISTRIBUTION(EventColors.FoodDistribution, R.string.event_type_food_distribuition), // + Food
+  ART_CONVENTION(
+      EventColors.ArtConvention, R.string.event_type_art_convention), // + Art, Literature
+  MANIFESTATION(
+      EventColors.Manifestation, R.string.event_type_manifestation), // + Culture, Socialising
+  BOARD_GAMES(EventColors.BoardGames, R.string.event_type_board_games), // + Gaming, Socialising
+  GROUP_STUDY(EventColors.GroupStudy, R.string.event_type_group_study), // + Science, Tech
+  OTHER(EventColors.Other, R.string.event_type_other)
```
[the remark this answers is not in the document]
oskar-codes: Also, for my own understanding, what do the comments indicate ? Is it the associated user interests ?

app/src/main/java/com/android/unio/ui/event/overlay/EventTypeOverlay.kt
```diff
@@ -0,0 +1,156 @@
...
+                    Surface(
+                        modifier = Modifier.sizeIn(maxHeight = 250.dp), color = Color.Transparent) {
+                          Column(modifier = Modifier.verticalScroll(scrollState)) {
+                            copiedTypes.forEachIndexed { index, pair ->
+
+                              // The row for each interest
+                              EventTypeOverlayRow(
+                                  pair,
+                                  onChange = {
+                                    isError.value = copiedTypes.count({ it.second.value }) > 3
```
oskar-codes, line 85: Consider extracting the value `3` into a constant to make the code more understandable and maintainable.

app/src/main/java/com/android/unio/ui/event/overlay/EventTypeOverlay.kt
```diff
@@ -0,0 +1,156 @@
...
+                                  })
+
+                              if (index != copiedTypes.size - 1) {
+                                HorizontalDivider(
+                                    modifier =
+                                        Modifier.testTag(
+                                            EventTypeOverlayTestTags.DIVIDER + "$index"))
```
oskar-codes, line 92: You can replace this by the shorter:
```kotlin
EventTypeOverlayTestTags.DIVIDER + index
```

oskar-codes (approved):
Phenomenal pull requests, your contributions are exemplary, keep it up 🚀

app/src/main/java/com/android/unio/model/event/Event.kt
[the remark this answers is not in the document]
AlouchLaBouche: I need to remove them, but yes it is the "potential matches" with the user interests that can be used with the custom recommendation system

app/src/main/java/com/android/unio/ui/event/overlay/EventTypeOverlay.kt
AlouchLaBouche: never

app/src/main/java/com/android/unio/ui/event/overlay/EventTypeOverlay.kt
AlouchLaBouche: good spot !

